PR #12, opened 2019-01-21
Separated build from test; Added Jenkinsfile

korobochka:
Integration with Jenkins


Jenkinsfile
```diff
@@ -0,0 +1,40 @@
+pipeline {
+    agent { label 'node' }
```
eroshenkoam: agent { docker 'npm:alpine' }

Issue #38, opened 2019-08-21
Allure-mocha does not show nested suites


just-boris:
#### I'm submitting a ... 
  - [ ] bug report
  - [x] feature request (some people may consider it a regression, it worked better in 1.x)
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

#### What is the current behavior?

In the report you see all tests on the flat level, without nesting

#### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem

```js
desctibe('a suite', () => {
   describe('child suite', () => {
      it('test one', () => {});
   });
   it('test two', () => {});
});
```

#### What is the expected behavior?

At the suites tab, I expect the following

* a suite
   * child suite
      * test one
  * test two

#### Please tell us about your environment:

| Allure version | 2.7.0 |
| --- | --- |
| Test framework | mocha@latest |
| Allure adaptor | allure-mocha@latest |
| Generate report using | allure-jenkins@latest |

#### Other information 

[//]: # (
. e.g. detailed explanation, stacktraces, related issues, suggestions 
. how to fix, links for us to have more context, eg. Stackoverflow, Gitter etc
)

PR #39, opened 2019-08-22
Refactor: add swappable writer to have in-memory results in testing

### LINKED ISSUE #37: Allure-mocha is not compatible with --check-leaks option
#### I'm submitting a ... 
  - [x] bug report
  - [ ] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

#### What is the current behavior?

When I use [`--check-leaks`](https://mochajs.org/#-check-leaks) flag, my tests fail.

#### If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem

It can be reproduced in the [demo package](https://github.com/sskorol/mocha-allure2-example). Add the `--check-leaks` flag [here](https://github.com/sskorol/mocha-allure2-example) and see the output:

```
0 passing (102ms)
  2 failing

  1) Authorization
       "before each" hook:
     Error: global leaks detected: allure, cache


  2) Authorization
       Sergey Korol should be able to sign:
     Error: No executable!
      at MochaAllureInterface.get [as currentExecutable] (node_modules/mocha-allure2-reporter/src/MochaAllureInterface.ts:124:13)
```

#### What is the expected behavior?

Reporter should be compatible with any mocha option.

#### Please tell us about your environment:

| Allure version | 2.7.0 |
| --- | --- |
| Test framework | mocha@5.2.0 |
| Allure adaptor | allure-mocha@latest |
| Generate report using | n/a |

just-boris:
### Context

This is a preparation to fix #37 and #38. Per discussion with @baev we need to update testing infra first. This PR does that

### Summary

* Updated `tsconfig.json`, remove the odd part about type roots, see individual commit for more details: https://github.com/allure-framework/allure-js/commit/1dc4ac3136645da8795f9cb8e6ebac178446762a

* Extracted `AllureWriter` interface from `AllureRuntime`, create two implementations, default and in-memory
* Updated all adapters to support the new writer option
* Replaced `GlobalInfoWriter` with new methods to support swappable writers

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-mocha/test/utils/index.ts
```diff
@@ -1,98 +1,27 @@
...
-const resultsDir = path.join(__dirname, "../../../out/allure-results");
-const resultsPollingInterval: number = 10;
 
-let results: any[] = [];
-
-export function runTests(...specs: string[]) {
+export function runTests(...specs: string[]): Promise<InMemoryAllureWriter> {
+  const writer = new InMemoryAllureWriter();
   const mocha = new Mocha();
+  mocha.reporter(MochaAllureReporter, { writer });
```
just-boris: Here are the changes in a nutshell. Now it is possible to provide a custom writer implementation via options and redirect all output somewhere else.

just-boris:
@baev what report? Since we are writing everything to memory, there are no files to generate a report

If you asked about running unit tests with the reporter enabled [here](https://github.com/allure-framework/allure-js/blob/master/packages/allure-mocha/test/mocha.opts#L4), it would require more work about adding and configuring [mocha-multi](https://www.npmjs.com/package/mocha-multi). Too much for the current PR

PR #104, opened 2020-03-06
Add dist/index.d.ts to package.json

### LINKED ISSUE #103: allure-jasmine allure-js-commons are missing types entry in package.json
https://github.com/allure-framework/allure-js/blob/master/packages/allure-js-commons/package.json
https://github.com/allure-framework/allure-js/blob/master/packages/allure-jasmine/package.json

Should have

```json
{
  "types": "dist/index.d.ts"
}
```

kirill-konshin:


just-boris:
Why is this needed? I am using `2.0.0-beta.6` with typescript and it resolves types just fine, without this change.

Moreover, this change points to a non-existing file `./dist/types.d.ts` which breaks typescript. You even had to work around this: https://github.com/allure-framework/allure-js/blob/master/packages/allure-mocha/test/utils/index.ts#L5

Can you give more context on what you were trying to do? I will try to help?

kirill-konshin:
Whoa, you’re right, I overlooked this typo, too much stuff was happening yesterday.

The reason for change is that if allure is used indirectly, through some intermediate lib, types can’t be traced by IDE without explicit types entry in package.json.

I will fix the mistake right now.

kirill-konshin:
https://github.com/allure-framework/allure-js/pull/109 done

PR #109, opened 2020-03-07
Fix #104

kirill-konshin:
Fixed typo & reverting ts-ignore


kirill-konshin:
Sure, thank you for your work.

Issue #50, opened 2019-09-03
allure-mocha - TypeError: Cannot read property 'epic' of undefined


fescobar:
I'm trying to do this configuration, but allure is undefined. Do you have any example calling allure method from mocha tests using TypeScript?
```
    "allure-mocha": "2.0.0-beta.6",
```

```
// es-modules
import { allure } from 'allure-mocha/runtime';
// or commonjs
const { allure } = require('allure-mocha/runtime');

it('is a test', () => {
  allure.epic('Some info');
});
```

```
TypeError: Cannot read property 'epic' of undefined
```
Thank you

just-boris:
It worked in my project.

Can you share your tech stack? What bundlers, compilers or other tools do you use?

fescobar:
I've added it a simple example. Install and execute the test (follow the README.md)
[example-mocha.zip](https://github.com/allure-framework/allure-js/files/3569557/example-mocha.zip)
Thank you so much

fescobar:
Hi @just-boris did you check my project? Thanks in advance.

fescobar:
@sseliverstov can you check the example to reproduce the error? 
https://github.com/allure-framework/allure-js/issues/50#issuecomment-527450874

Thanks

ryparker:
Same issue here. Is typescript a requirement?

ryparker:
Can we get an update on this? This is blocking my ability to use this library. I've even cloned the allure-mocha example just to test this in a typescript environment and it is also giving this same error. Is the allure-mocha library not meant to be used outside of typescript??

dzotovwork:
the same problem

dzotovwork:
![image](https://user-images.githubusercontent.com/43908260/80400690-d3b3d600-88c3-11ea-9e83-0d58ed99cd13.png)
it is master branch from https://github.com/sskorol/mocha-allure2-example
just add     "allure-mocha": "^2.0.0-beta.6", in devdeps and run `yarn` then `test` script from package.json and try allure.createStep and catch error :(

ostul:
I have the same problem

dzotovwork:
Total, would like a working example with mocha-allure, ts-test-decorators or @ testdeck / mocha, it would be ideal with multi-threading) I think everyone will be very grateful)

sskorol:
I'll take a look at it. But probably only by the end of this week.

sskorol:
[...]

Issue #48, opened 2019-08-28
Support before and after hooks


just-boris:
#### I'm submitting a ... 
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

#### What is the current behavior?

Currently, all steps and attachments in `before` and `after` sections are getting ignored. We should provide an ability to record them to the report.

#### What is the expected behavior?

this code should work, attachment added to the report.

```
before(() => {
   allure.attachment('from before', 'hello!', ContentType.TEXT);
})
```
#### Please tell us about your environment:

| Allure version | 2.7.0 |
| --- | --- |
| Test framework | jest@latest |
| Allure adaptor | allure-mocha@latest |
| Generate report using | allure-jenkins@latest |

#### Other information 

[//]: # (
. e.g. detailed explanation, stacktraces, related issues, suggestions 
. how to fix, links for us to have more context, eg. Stackoverflow, Gitter etc
)

guitartsword:
I want to add a file with the current test on an afterEach
```
afterEach(function () {
  allure.attachment('File att', 'File Data', ContentType.TEXT);
  allure.testAttachment('File tatt', 'File Data', ContentType.TEXT);
  allure.createAttachment('File catt', 'File Data', ContentType.TEXT);
});
```

Opening the `allure-results` folder I can see the attachment was created, but the test doesn't have the files. I guess the problem is that the `afterEach` hook is treated as another test case and this test case is not generated causing the attachment that created not to be link to a test case.

What we want is something like the following:

- **before Hook:** should have all the allure config in the first test case from the describe block
- **after Hook:** same as before hook, only that the allure calls should be linked to the last test case
- **beforeEach & afterEach Hooks:** This should be linked to the test that this hook is related


```
describe('Suite Name', function () {
  before(function () {
    allure.tag('first');
    allure.tag('before');
  });
  after(function () {
    allure.tag('last');
    allure.tag('after');
  });
  beforeEach(function () {
    allure.tag('beforeEach');
  });
  afterEach(function () {
    allure.tag('afterEach');
  });
  it('Test One', function () {
    allure.issue('1', 'https://www.example.com?issue=1');
  });
  it('Test Two', function () {
    allure.issue('2', 'https://www.example.com?issue=2');
  });
  it('Test Three', function () {
    allure.issue('3', 'https://www.example.com?issue=3');
  });
});
```

Notes:
- BeforeEach Hook is working
[...]

PR #138, opened 2020-05-03
feat(mocha): revise current allure-mocha implementation

### LINKED ISSUE #137: Revise allure-mocha
[//]: # (
. Note: for support questions, please use Stackoverflow or Gitter**. 
. This repository's issues are reserved for feature requests and bug reports.
.
. In case of any problems with Allure Jenkins plugin** please use the following repository 
. to create an issue: https://github.com/jenkinsci/allure-plugin/issues
.
. Make sure you have a clear name for your issue. The name should start with a capital 
. letter and no dot is required in the end of the sentence. An example of good issue names:
.
. - The report is broken in IE11
. - Add an ability to disable default plugins
. - Support emoji in test descriptions
)

#### I'm submitting a ... 
  - [ ] bug report
  - [x] feature request
  - [ ] support request => Please do not submit support request here, see note at the top of this template.

#### What is the current behavior?
Current allure-mocha version seems a bit outdated:
- dependencies are outdated;
- hooks are not correctly handled;
- mocha-typescript is deprecated;
- it's hard to apply decorators with such a tricky exports;
- some new features like "epic" are uncovered.

#### What is the motivation / use case for changing the behavior?
It's hard to reuse current allure-mocha implementation within other frameworks or libraries. So we need to consider the following updates:
- update package dependencies to the latest versions;
- update typescript dependency in a root module, as re-exporting issue is already fixed;
- replace deprecated `mocha-typescript` with `@testdeck/mocha`;
- add before/after hooks handlers;
- add Allure container to the root `allure-js-commons` for easier decorators' implementation;
- revise module exports;
- cover "epic" feature.

sskorol:
As it's hard to reuse current allure-mocha implementation within other frameworks or libraries, there were made the following modifications:
- updated package dependencies to the latest versions;
- replaced deprecated `mocha-typescript` with `@testdeck/mocha`;
- added before/after hooks handlers (note that before hook won't work as expected until [mocha issue](https://github.com/mochajs/mocha/issues/4266) is fixed or we revise the current contexts' switching approach);
- covered `epic` feature;
- updated all the tests to support `@testdeck/mocha`.

Fixes #137, #50, #48 (partially), #7 (baseline).

Note that I've already tested it locally with decorators and standalone reporter examples. Current version is correctly injected as a default reporter into Mocha context. When we release a new decorators' module (will raise a separate PR), it'll be possible to wrap any reporter's instance for further decorators' usage.

Sample test with all the features:
```typescript
// ...
import { suite, test } from '@testdeck/mocha'
import { ContentType, Severity } from 'allure-js-commons'
import { allure, MochaAllure } from 'allure-mocha'
import { data, description, epic, feature, issue, owner, severity, story, tag, testCaseId, decorate } from 'allure-decorators'

@suite
class Authorization {
  public static testData = (): User => {
    return User.dummy()
  }

  @issue('1')
  @testCaseId('5')
  @severity(Severity.BLOCKER)
  @epic('User Authentication')
  @feature('Login')
  @story('Positive authorization')
  @owner('skorol')
  @tag('smoke')
  @description('Basic authorization test.')
  @data(Authorization.testData)
  @data.naming(user => `${user} should be able to sign`)
  @test
  public userShouldBeAbleToSignIn(user: User) {
    open(LoginPage)
      .loginWith(user)
      .select(ProfilePage)

    verifyThat(atProfilePage)
      .fullNameIs(user.fullName)
      .usernameIs(user.username)
  }

  public before() {
    decorate<MochaAllure>(allure)
    setUpCache()
  }

  public after() {
    this.allure.attachment('Test attachment', 'test attachment content', ContentType.TEXT)
    cleanUpCache()
  }
}

```

The result on the recent version looks like the following:
image

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-mocha/package.json
+++ b/packages/allure-mocha/package.json
@@ -25,24 +25,24 @@
     "lint-fix": "eslint ./src ./test ./index.ts --ext .ts --fix"
   },
   "devDependencies": {
-    "@types/chai": "^4.1.7",
+    "@testdeck/mocha": "0.0.10",
+    "@types/chai": "^4.2.11",
     "@types/mocha": "^7.0.2",
     "chai": "^4.2.0",
-    "codecov": "^3.3.0",
-    "dotenv": "^8.0.0",
-    "fs-jetpack": "^2.2.2",
-    "mocha": "^6.2.0",
-    "mocha-multi": "^1.1.1",
-    "mocha-typescript": "^1.1.17",
-    "nyc": "^15.0.0",
-    "prettier": "^1.17.0",
-    "source-map-support": "^0.5.12"
+    "codecov": "^3.6.5",
+    "dotenv": "^8.2.0",
+    "fs-jetpack": "^2.2.3",
+    "mocha": "^7.1.2",
+    "mocha-multi": "^1.1.3",
+    "nyc": "^15.0.1",
+    "prettier": "^2.0.5",
+    "source-map-support": "^0.5.19"
   },
   "dependencies": {
     "allure-js-commons": "^2.0.0-beta.6"
   },
   "peerDependencies": {
-    "mocha": "^6.2.0"
+    "mocha": ">=6.2.x"
   },
   "nyc": {
     "check-coverage": true,
@@ -53,8 +53,10 @@
     "extension": [
       ".ts"
     ],
-    "include": [
-      "dist/src/**/*.js"
+    "exclude": [
+      "test/**/*.*",
+      "**/*.d.ts",
+      "runtime.js"
     ],
     "reporter": [
       "lcov",
--- a/packages/allure-mocha/runtime.d.ts
+++ b/packages/allure-mocha/runtime.d.ts
@@ -1,1 +1,3 @@
-export { allure } from "./dist/MochaAllureReporter";
+export { MochaAllure } from "./dist/MochaAllure";
+export { MochaAllureReporter } from "./dist/MochaAllureReporter";
+export { allure } from './dist/MochaAllureReporter';
--- a/packages/allure-mocha/runtime.js
+++ b/packages/allure-mocha/runtime.js
@@ -5,11 +5,27 @@
 // https://babeljs.io/repl#?code_lz=KYDwDg9gTgLgBAbwIYBsUFcrAL5wGZQQC2cA5AHQD0AJgJYDOMl9UAxpQLISsAWSAgmkzAASsEixgUUkA
 "use strict";
 
-const reporter = require("./dist/MochaAllureReporter");
+const _MochaAllure = require("./dist/MochaAllure");
+const _MochaAllureReporter = require("./dist/MochaAllureReporter");
 
-Object.defineProperty(exports, "__esModule", { value: true });
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+Object.defineProperty(module.exports, "MochaAllure", {
+  enumerable: true,
+  get: function () {
+    return _MochaAllure.MochaAllure;
+  }
+});
 Object.defineProperty(module.exports, "allure", {
-  get() {
-    return reporter.allure;
+  enumerable: true,
+  get: function () {
+    return _MochaAllureReporter.allure;
+  }
+});
+Object.defineProperty(module.exports, "MochaAllureReporter", {
+  enumerable: true,
+  get: function () {
+    return _MochaAllureReporter.MochaAllureReporter;
   }
 });
--- a/packages/allure-mocha/src/AllureReporter.ts
+++ b/packages/allure-mocha/src/AllureReporter.ts
@@ -4,37 +4,34 @@ import {
   AllureStep,
   AllureTest,
   ContentType,
+  ExecutableItemWrapper,
   LabelName,
   Stage,
   Status,
   StatusDetails
 } from "allure-js-commons";
 import { createHash } from "crypto";
-import { MochaAllureInterface } from "./MochaAllureInterface";
+import { MochaAllure } from "./MochaAllure";
 
 export class AllureReporter {
   private suites: AllureGroup[] = [];
   private steps: AllureStep[] = [];
   private runningTest: AllureTest | null = null;
+  public currentExecutable: ExecutableItemWrapper | null = null;
 
-  constructor(private runtime: AllureRuntime) {}
+  constructor(private readonly allureRuntime: AllureRuntime) {
+  }
 
-  public getInterface(): MochaAllureInterface {
-    return new MochaAllureInterface(this, this.runtime);
+  public getImplementation(): MochaAllure {
+    return new MochaAllure(this, this.allureRuntime);
   }
 
   get currentSuite(): AllureGroup | null {
-    if (this.suites.length === 0) {
-      return null;
-    }
-    return this.suites[this.suites.length - 1];
+    return this.suites.length > 0 ? this.suites[this.suites.length - 1] : null;
   }
 
   get currentStep(): AllureStep | null {
-    if (this.steps.length > 0) {
-      return this.steps[this.steps.length - 1];
-    }
-    return null;
+    return this.steps.length > 0 ? this.steps[this.steps.length - 1] : null;
   }
 
   get currentTest(): AllureTest | null {
@@ -45,13 +42,13 @@ export class AllureReporter {
     this.runningTest = test;
   }
 
-  public startSuite(suiteName: string) {
-    const scope = this.currentSuite || this.runtime;
-    const suite = scope.startGroup(suiteName || "Global");
+  public startSuite(suiteName: string): void {
+    const scope: AllureGroup | AllureRuntime = this.currentSuite || this.allureRuntime;
+    const suite: AllureGroup = scope.startGroup(suiteName || "Global");
     this.pushSuite(suite);
   }
 
-  public endSuite() {
+  public endSuite(): void {
     if (this.currentSuite !== null) {
       if (this.currentStep !== null) {
         this.currentStep.endStep();
@@ -61,7 +58,30 @@ export class AllureReporter {
     }
   }
 
-  public startCase(test: Mocha.Test) {
+  public startHook(title: string): void {
+    const suite: AllureGroup | null = this.currentSuite;
+
+    if (suite && title && title.indexOf("before") !== -1) {
+      this.currentExecutable = suite.addBefore();
+    } else if (suite && title && title.indexOf("after") !== -1) {
+      this.currentExecutable = suite.addAfter();
+    }
+  }
+
+  public endHook(error?: Error): void {
+    if (this.currentExecutable) {
+      this.currentExecutable.stage = Stage.FINISHED;
+
+      if (error) {
+        this.currentExecutable.status = Status.FAILED;
+        this.currentExecutable.statusDetails = { message: error.message, trace: error.stack };
+      } else {
+        this.currentExecutable.status = Status.PASSED;
+      }
+    }
+  }
+
+  public startCase(test: Mocha.Test): void {
     if (this.currentSuite === null) {
       throw new Error("No active suite");
     }
@@ -87,19 +107,19 @@ export class AllureReporter {
     }
   }
 
-  public passTestCase(test: Mocha.Test) {
+  public passTestCase(test: Mocha.Test): void {
     if (this.currentTest === null) {
       this.startCase(test);
     }
     this.endTest(Status.PASSED);
   }
 
-  public pendingTestCase(test: Mocha.Test) {
+  public pendingTestCase(test: Mocha.Test): void {
     this.startCase(test);
     this.endTest(Status.SKIPPED, { message: "Test ignored" });
   }
 
-  public failTestCase(test: Mocha.Test, error: Error) {
+  public failTestCase(test: Mocha.Test, error: Error): void {
     if (this.currentTest === null) {
       this.startCase(test);
     } else {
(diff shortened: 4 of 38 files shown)


packages/allure-mocha/package.json
```diff
@@ -25,24 +25,24 @@
...
+    "nyc": "^15.0.1",
+    "prettier": "^2.0.5",
+    "source-map-support": "^0.5.19"
   },
   "dependencies": {
     "allure-js-commons": "^2.0.0-beta.6"
   },
   "peerDependencies": {
-    "mocha": "^6.2.0"
+    "mocha": "^7.1.2"
```
just-boris, line 45: Is there any particular reason why this package can't support Mocha 6 and 7?

packages/allure-mocha/src/AllureReporter.ts
```diff
@@ -45,13 +42,21 @@ export class AllureReporter {
...
-  public startSuite(suiteName: string) {
-    const scope = this.currentSuite || this.runtime;
-    const suite = scope.startGroup(suiteName || "Global");
+  get currentExecutable(): ExecutableItemWrapper | null {
+    return this.runningExecutable;
+  }
+
+  set currentExecutable(executable: ExecutableItemWrapper | null) {
+    this.runningExecutable = executable;
+  }
```
just-boris, line 51: Can this be just a public property?

packages/allure-mocha/src/index.ts
```diff
@@ -1,3 +1,8 @@
 import { MochaAllureReporter } from "./MochaAllureReporter";
+import { AllureContainer } from "allure-js-commons";
+import { MochaAllure } from "./MochaAllure";
 
-export = MochaAllureReporter;
```
just-boris, line 3: What's the reason to change `export = ` to `module.exports =`? It is the same output in Javascript, but `export = ` also generates proper .d.ts

packages/allure-mocha/test/fixtures/specs/attachment.ts
```diff
@@ -1,20 +1,23 @@
 import { ContentType } from "allure-js-commons";
-import { suite, test } from "mocha-typescript";
-import { allure } from "../../../runtime";
```
just-boris, line 3: Can we keep this import instead of calling `MochaAllureReporter.allureContainer.getImplementation()`.

I see that you export `allureContainer`, so should be possible, and the diff below will be smaller

packages/allure-mocha/test/runner.ts
```diff
@@ -5,16 +5,9 @@ import path from "path";
...
-  interface InterfaceContributions {
-    "mocha-typescript": never;
-  }
-}
-
 const mocha = new Mocha({
-  ui: "mocha-typescript",
   timeout: 16000,
-  reporter: "mocha-multi",
+  reporter: "list",
```
just-boris, line 10: Mocha-multi was used here to generate a demo report from the tests. Can we keep this?

packages/allure-mocha/runtime.js
```diff
@@ -1,15 +0,0 @@
...
-"use strict";
-
-const reporter = require("./dist/MochaAllureReporter");
-
-Object.defineProperty(exports, "__esModule", { value: true });
-Object.defineProperty(module.exports, "allure", {
-  get() {
-    return reporter.allure;
-  }
-});
```
just-boris, line 15: FWIW, the fix will be released in Typescript 3.9 which is happening very soon which eliminates this hack. Maybe we could have kept this separate import for runtime.

sskorol:
@just-boris thanks for the comments, I'll take a look at runtime again. I had some issues while importing required dependencies from this module while refactoring. Perhaps, I missed something. Need to double check.

packages/allure-mocha/src/AllureReporter.ts
sskorol: Yes, will change.

packages/allure-mocha/src/index.ts
sskorol: The main reason is in additional exports. TS doesn't allow to do it if we have anything else here:
image

packages/allure-mocha/package.json
sskorol: Will change to `>=6.2.x`

packages/allure-mocha/test/runner.ts
sskorol: I did it just because it uses only root specs in report. So we don't see any specific features under test (fixtures). But no problem, will revert.

packages/allure-mocha/src/index.ts
just-boris: Right, because you can't mix common js and es exports in the same file. Basically the last `module.exports =` assignment erases all exports above.

This is the reason why the current master branch uses a separate `runtime` file to get around this.

sskorol:
@just-boris tried to add runtime back with the following updates: 

*runtime.d.ts*
```typescript
export { MochaAllure, MochaAllureReporter, allureContainer } from "./dist";
```
*runtime.js* (generated via Babel link)
```js
"use strict";

const root = require("./dist");

Object.defineProperty(exports, "__esModule", {
  value: true
});
Object.defineProperty(exports, "allureContainer", {
  enumerable: true,
  get: function () {
    return root.allureContainer;
  }
});
Object.defineProperty(exports, "MochaAllureReporter", {
  enumerable: true,
  get: function () {
    return root.MochaAllureReporter;
  }
});
Object.defineProperty(exports, "MochaAllure", {
  enumerable: true,
  get: function () {
    return root.MochaAllure;
  }
});
```
Also changed tests' imports like the following:
```typescript
import { MochaAllure, allureContainer } from "../../../";

@suite
class AttachmentSubSuite {
  private readonly allure: MochaAllure = allureContainer.getImplementation();
```
And *util/index.ts*
```typescript
import { MochaAllureReporter } from "../../";
```
If I leave it this way, a runner fails the upstream suite with "Test not found" error. Obviously, the reason is in a last snippet. Not sure why, but the shortened path (neither "../../" nor "../../dist") is not resolved correctly. However, when I change it to "../../dist/MochaAllureReporter", all the tests are passing. But in this case TS shows the warning that this full pass can be shortened. Any idea what I'm missing here?

packages/allure-mocha/src/index.ts
sskorol: So if a leave only a reporter export here, then how would I export a static `allureContainer` from within `MochaAllureReporter`? Or I'd need to move it outside a class the way it was implemented before, and then export via getter in runtime.js?

sskorol: Ok, figured it out. Will push updates soon.

packages/allure-mocha/test/fixtures/specs/attachment.ts
sskorol: Yes, will change it to the old approach.

packages/allure-mocha/runtime.js
sskorol: Thought it was already released, will update.

sskorol:
@just-boris made all the fixes. Checked locally with decorators and high-level examples. Everything works as expected. Thanks.
Let me know if there's anything else. Otherwise, will squash commits.

just-boris:
Why did you wrap from `allure` with an extra container? What was missing in the previous approach with exporting the `allure` variable directly?

sskorol:
@just-boris the initial intention was adding a generic container for further decorators' module implementation, which is report-depended. But when you asked the above question, I realized that current approach is not generic at all. So in a latest commit I've moved AllureContainer init into commons module. Now we can provide the implementation on fly. However, it requires additional casting while implementation retrieval.

sskorol:
@just-boris so the main idea is the following. First, we init and export a container within `allure-js-commons` module:
```typescript
export const allureContainer: AllureContainer<Allure> = new AllureContainer();
```
This allows any reporter to add a required `Allure` implementation in runtime:
```typescript
export class MochaAllureReporter extends Mocha.reporters.Base {
  private coreReporter: AllureReporter;

  constructor(readonly runner: Mocha.Runner, readonly opts: Mocha.MochaOptions) {
    super(runner, opts);
    // ...
    allureContainer.setImplementation(this.coreReporter.getImplementation());
```
To be able to access this implementation in any other package, we just need to apply a reporter (e.g. via `reporter` option in Mocha), and then import the following dependencies:
```typescript
import { allureContainer, Allure } from 'allure-js-commons'
// If we need a particular type
import { MochaAllure } from 'allure-mocha/runtime'
```
Now we can retrieve the required implementation via one of the following:
```typescript
function getAllure(): MochaAllure {
  // Either directly (if we don't need report-specific API)
  const allure: Allure = allureContainer.getImplementation()
  // Or via casting to the required report type
  const allure: MochaAllure = allureContainer.getImplementation() as MochaAllure

  if (!allure) {
    throw new Error('Unable to find MochaAllure implementation')
  }

  return allure
}
```

just-boris:
How does having a container help with anything? In the current `2.0.0-beta.7` you can do the following:

```js
import { allure } from 'allure-mocha/runtime';

it('is a test', () => {
  allure.createAttachment('something', 'something', ContentType.TXT);
})
```

After this PR, there will be an extra line:

```js
import { allureContainer } from 'allure-mocha/runtime';

// you can't place this line here, because implementation is not ready yet and you will get undefined
// const allure = allureContainer.getImplementation()

it('is a test', () => {
   // works only inside of a test or before/after blocks
  const allure = allureContainer. getImplementation()
  allure.createAttachment('something', 'something', ContentType.TXT);
})
```

This line also has a confusing behavior, similar to https://github.com/allure-framework/allure-js/issues/50, where the object is undefined before the reporter is ready.

I guess, this container is somehow useful for the test-decorators package, but I don't see how. What is not possible with `2.0.0-beta.7` API, which will be achievable after the PR? I can try finding you a solution that does not compromise vanilla mocha users.

sskorol:
Well, imagine we want to add a decorators' package into this repository (or if any third-party wants to provide such support). Decorators' module should be supplied with Allure instance to wrap the required API, respecting the context. Each reporter's package extends Allure and provides an overridden behaviour for e.g. steps. Assuming the fact that Allure implementation could be set only in runtime (when reporter is initiated), we need to find a way to supply a report-independent implementation to a dedicated decorators' module. Recent version (2.0.0-beta.7) is Mocha-specific. An introduced container (in commons package) abstracts away from Mocha or any other framework. It defines a generic contract which assumes setting only Allure child classes in runtime, so that independently from a reporter that is currently used as a dependency, we can work with a generic container from commons module instead of a report-specific implementation. Decorators' module must be independent. We've discussed this issue with @korobochka before. And there were 2 options in mind: DI or generic container. But if you have any other suggestions, I'm open to discuss it in e.g. Allure slack chat.

Regarding visibility scope: yes, it becomes context-dependent. But isn't it totally fine when the entire Allure is an event-driven framework? We access a corresponding reference only when we need it.

just-boris:
So far, I see two topics here.

1. With this change, using `test-decorators` becomes a requirement to use Allure. If they don't use decorators, they need to write boilerplate code in any file. Can we find a way to make test-decorators work, without impacting vanilla JS users?

2. Moving the global state container to the commons. Global states are usually an anti-pattern and should be avoided. Jasmine and Cucumber have convenient programmatic API, so they do not require global state. It is an unfortunate limitation of Mocha. 

As I understand, you want to make test-decorators package generic for any framework, via direct dependency on allure-js-commons. How about this API instead:

```javascript
import {setAllure} from 'test-decorators';
import {allure} from 'allure-mocha';

setAllure(allure);
```

1. This should work with any framework, without any changes in Allure
1. Explicit is better than implicit, customers clearly see what they do
1. No issues with conflict dependencies

sskorol:
You mean delegating `allure` instance setting API (for decorators module) to the client side instead of trying to incapsulate it into Allure internals? Sounds reasonable. Give me couple of days to experiment with this approach.

sskorol:
@just-boris ok, I tested it locally and seems like it'll work. Reverted updates related to container and squashed the commits.

packages/allure-mocha/runtime.d.ts
```diff
@@ -1 +1,3 @@
-export { allure } from "./dist/MochaAllureReporter";
+export { MochaAllure } from "./dist/MochaAllure";
+export { MochaAllureReporter } from "./dist/MochaAllureReporter";
```
just-boris, line 2: Should these values be exported from `index.ts` instead of runtime?

sskorol: As in `index.ts` we have the following:
```
export = MochaAllureReporter;
```
TS doesn't allow us putting any other exports there.

packages/allure-mocha/test/utils/index.ts
```diff
@@ -1,7 +1,7 @@
 import jetpack = require("fs-jetpack");
 import * as path from "path";
 import Mocha from "mocha";
-import MochaAllureReporter from "../..";
+import { MochaAllureReporter } from "../../dist/MochaAllureReporter";
```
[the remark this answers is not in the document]
sskorol: Yes, reverted.

packages/allure-mocha/runtime.d.ts
just-boris: Understood, let's keep it this way

just-boris:
@baev @sseliverstov to have a final look and merge.

sskorol:
@baev @sseliverstov folks, any updates?

sskorol:
@just-boris do you have their contacts? With such response rate, it's easier to maintain it in my own repo. :)

PR #221, opened 2020-11-13
feat: Support custom attachment content types

just-boris:
This is the Javascript integration for this feature: https://github.com/allure-framework/allure2/pull/1145

Currently only pre-defined content types are supported, this PR adds custom types:

```js
allure.attachment('screen diff', screenDiffJson, {
   contentType: 'application/vnd.allure.image.diff',
   fileExtension: 'json' // no auto-detect, because `file-type` npm module has async API, but our method is sync
})
```


just-boris:
Could @baev or @sseliverstov check this PR? it is needed to leverage the new screen diff improvements in JS frameworks

baev (approved):

PR #298, opened 2021-07-17
fix jasmine type definitions

just-boris:
### Context

Currently the build in the `master` branch fails, because of changes in the `@types/jasmine` dependency. This PR fixes that

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


just-boris:
This should not have happened in the first place. The project should not put `package-lock.json` files into `.gitignore`, this file protects against this kind of surprise changes

baev (approved):

PR #323, opened 2021-08-03
Build improvements

baev:
### Misc

* Add global eslint config
* Add global prettier
* Husky to run prettier precommit hooks
* Fix code style according to meet new source code requirements

### Breaking

* Remove few deprecations (No more interfaces with I prefix)
    `AllureConfig` was removed
    `IAllureConfig` was renamed to `AllureConfig`
    `AllureWriter` was renamed to `FileSystemAllureWriter`
    `IAllureWriter` was renamed to `AllureWriter`

* All API methods now has return type specified

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


sseliverstov (approved):

packages/allure-jest/test/default-runner/simple.test.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line no-undef
```
just-boris, line 1: Should be fixed with `/*eslint-env jest*/`?

.eslintrc.js
```diff
@@ -0,0 +1,256 @@
...
+    "one-var": ["off", "never"],
+    "padding-line-between-statements": [
+      "off",
+      {
+        blankLine: "always",
+        prev: "*",
+        next: "return",
+      },
+    ],
+    "prefer-arrow/prefer-arrow-functions": "error",
```
just-boris, line 232: Not sure about this one: `const demo = () => {}` is 21 character, but `function demo() {}` is 18. Not sure why someone would prefer a longer form.

There is a built-in "[prefer arrow callback](https://eslint.org/docs/rules/prefer-arrow-callback)" rule which makes sense, but using arrow for all is too much imo.

just-boris (commented):
minor suggestions

packages/allure-jest/test/default-runner/simple.test.ts
baev: done

.eslintrc.js
baev: it is alignment with our internal code style, so I like to keep it

.github/workflows/build.yml
```diff
@@ -26,9 +26,18 @@ jobs:
           npm run prepare
         env:
           CI: true
-      - name: npm build, test
+      - name: npm build
         run: |
           npm run build
+        env:
```
mxschmitt, line 32: FYI, `CI` is always set to `true`: https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables

PR #327, opened 2021-08-17
Fix lerna

baev:
Use relative npm dependenencies and add missing package locks.

#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/package.json
+++ b/packages/allure-cucumberjs/package.json
@@ -38,6 +38,6 @@
     "verror": "^1.10.0"
   },
   "dependencies": {
-    "allure-js-commons": "^2.0.0-beta.11"
+    "allure-js-commons": "file:../allure-js-commons"
   }
 }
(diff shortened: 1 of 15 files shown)


packages/allure-cucumberjs/package.json
```diff
@@ -38,6 +38,6 @@
     "verror": "^1.10.0"
   },
   "dependencies": {
-    "allure-js-commons": "^2.0.0-beta.11"
+    "allure-js-commons": "file:../allure-js-commons"
```
just-boris, line 41: What happens when you publish this to npm? It will not work, if it stays unchanged.

Could you test it by running `npm publish --dry-run`?

baev: I'm pretty sure lerna will replace it

eroshenkoam (approved):

packages/allure-cucumberjs/package.json
just-boris: Your bet. Good luck!

PR #329, opened 2021-08-17
Fix playwright path attachments

### LINKED ISSUE #326: [allure-playwright]: Artifacts have fullpath in allure results
Mirroring https://github.com/microsoft/playwright/issues/8079

@baev should I be calling writeAttachment with relative file name? If yes, relative to what?

baev:
#### Context

fixes #326

the change also intruduces new API methods:

```js
AllureRuntime#writeAttachmentFromPath(string, PathLike);
AllureWriter#writeAttachmentFromPath(string, PathLike);
```

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/src/writers/FileSystemAllureWriter.ts
+++ b/packages/allure-js-commons/src/writers/FileSystemAllureWriter.ts
@@ -22,7 +22,7 @@ export class FileSystemAllureWriter implements AllureWriter {
     writeFileSync(path, content);
   }
 
-  writeAttachmentPath(targetFileName: string, sourceFilePath: PathLike): void {
+  writeAttachmentFromPath(targetFileName: string, sourceFilePath: PathLike): void {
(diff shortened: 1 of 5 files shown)


packages/allure-js-commons/src/writers/FileSystemAllureWriter.ts
```diff
@@ -22,6 +22,11 @@ export class FileSystemAllureWriter implements AllureWriter {
     writeFileSync(path, content);
   }
 
+  writeAttachmentPath(targetFileName: string, sourceFilePath: PathLike): void {
```
just-boris, line 25: Reading the method name I expect that it writes the path somewhere, but not the actual file.

How about renaming it to `writeAttachmentFromPath`?

just-boris (approved):

PR #344, opened 2021-08-31
Migration to yarn workspaces

baev:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-js-commons/src/writers/InMemoryAllureWriter.ts
```diff
@@ -2,6 +2,12 @@ import { PathLike, readFileSync } from "fs";
 import { Category, TestResult, TestResultContainer } from "../model";
 import { AllureWriter } from "./AllureWriter";
 
+export interface AllureResults {
```
just-boris, line 5: What happened here? How did this work before

just-boris (approved):
One non-blocking question

packages/allure-js-commons/src/writers/InMemoryAllureWriter.ts
baev: that is additional interface, that can be used in together with `InMemoryAllureWriter`. This is part of the future changes I partially picked from other branch (and this interface is added by mistake actually). The idea is to encapsulate writer (return `AllureResults`  instead of `InMemoryAllureWriter`):

```js
describe("Allure XYZ integration tests", () => {
  it("should set status", () => {
    const results: AllureResults = runXyzSpec("testdata/status_spec.ts"); 
    const [ testResult ] = results;
    expect(testResult.status).to.be.eq(Status.PASSED)
  })
})

// xyz_test_utils.ts
const runXyzSpec = (specPath: PathLike): AllureResults => {
   const writer = new InMemoryResultsWriter();
   const reporter = new AllureXyz(writer);
  
   XYZ.runTests(specPath, { reporter });

   return { tests: writer.tests, groups: writer.groups, attachments: writer.attachments };
}

```

baev: I'm not feel like sending separate merge request with this change, so I'll merge it

just-boris: ok, as you wish

PR #349, opened 2021-09-01
fix(playwright): do not throw if cannot copy attachment from given path

yury-s:
### Context
Some of the attachments may be deleted bu the runner (e.g. `video: on, preserveOutput: 'failures-only'` test config) or clients could add non-existing path as attachment.

Fixes https://github.com/microsoft/playwright/issues/8611

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-playwright/src/index.ts
+++ b/packages/allure-playwright/src/index.ts
@@ -92,9 +92,14 @@ class AllureReporter implements Reporter {
               continue;
             }
 
-            const fileName = attachment.body
-              ? runtime.writeAttachment(attachment.body, attachment.contentType)
-              : runtime.writeAttachmentFromPath(attachment.path!, attachment.contentType);
+            let fileName;
+            try {
+              fileName = attachment.body
+                ? runtime.writeAttachment(attachment.body, attachment.contentType)
+                : runtime.writeAttachmentFromPath(attachment.path!, attachment.contentType);
+            } catch(e) {
+              continue;
+            }
             allureTest.addAttachment(attachment.name, attachment.contentType, fileName);
             if (attachment.name === "diff") {
               allureTest.addLabel("testType", "screenshotDiff");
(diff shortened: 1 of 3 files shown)


packages/allure-playwright/src/index.ts
```diff
@@ -92,9 +92,14 @@ class AllureReporter implements Reporter {
...
             }
 
-            const fileName = attachment.body
-              ? runtime.writeAttachment(attachment.body, attachment.contentType)
-              : runtime.writeAttachmentFromPath(attachment.path!, attachment.contentType);
+            let fileName;
+            try {
+              fileName = attachment.body
+                ? runtime.writeAttachment(attachment.body, attachment.contentType)
+                : runtime.writeAttachmentFromPath(attachment.path!, attachment.contentType);
```
baev, line 99: can we add additional validation that file by specified `attachment.path` is exists before actually trying to copy it?

yury-s: I can add that but I feel like there may be more issues, e.g. not enough permissions to copy the file etc. Do you think simple existence check would be good enough?

baev: I’d say let’s check if filePath is present and file exits by that path, and if it doesn’t just skip the attachment. And in case of exception (on existence check or clone) we should report error to user (insufficient permissions is a good case, probable something wrong with user’s setup)

yury-s: Makes sense. Update the code to silently continue when the file doesn't exist and throw on all other errors as before. Long-term we'll need an interface in the playwright test runner for communicating back such exceptions in the reporter as the users usually don't expect their test suite to crash in the middle because of an exception in the reporter. Currently we don't have such API.

baev (approved):

baev:
@yury-s thanks!

PR #408, opened 2021-11-18
Add labels and links to playwright via attachments #332  #352 #341

vovsemenv:
### Context
Added allure metainfo to playwright via attachments.
resolve #332 #352 #341

#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests

[cla]: https://cla-assistant.io/accept/allure-framework/allure2

PR #412, opened 2021-12-03
Generate Allure results in real-time

dtinth:
### Context
The original implementation

- Generates Allure tests results data only when test is finished.
- Has some invalid methods (e.g. `onTimeout`).

This PR upgrades Playwright Test to latest version to reflect latest Playwright API. All tests are passing.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-playwright/src/index.ts
+++ b/packages/allure-playwright/src/index.ts
@@ -66,7 +66,7 @@ class AllureReporter implements Reporter {
 
   onTestBegin(test: TestCase): void {
     const suite = test.parent;
-    const group = this.getAllureGroup(suite);
+    const group = this.ensureAllureGroupCreated(suite);
     const allureTest = group.startTest(test.title);
     allureTest.addLabel(LabelName.LANGUAGE, "JavaScript");
     allureTest.addLabel(LabelName.FRAMEWORK, "Playwright");
@@ -90,7 +90,7 @@ class AllureReporter implements Reporter {
     if (!allureTest) {
       return;
     }
-    this.getAllureStep(step, allureTest);
+    this.ensureAllureStepCreated(step, allureTest);
   }
 
   onStepEnd(_test: TestCase, _result: TestResult, step: TestStep): void {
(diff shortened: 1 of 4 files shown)


packages/allure-playwright/src/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
just-boris, line 1: why is this needed? There are `eslint-disable-line` below

packages/allure-playwright/src/index.ts
```diff
@@ -17,130 +18,160 @@
...
+        console.log(JSON.stringify({ error: (e as Error).stack || String(e) })); // eslint-disable-line no-console
       }
     }
   }
+
+  private getAllureGroup(suite: Suite): AllureGroup {
+    let group = this.allureGroupCache.get(suite);
+    if (!group) {
+      const parent = suite.parent ? this.getAllureGroup(suite.parent) : this.allureRuntime;
+      group = parent.startGroup(suite.title);
```
just-boris, line 160: It is not expected that `get*` methods have any side effects. Better to call it `ensureGroupStarted`

packages/allure-playwright/src/index.ts
```diff
@@ -17,130 +18,160 @@
...
+      this.allureGroupCache.set(suite, group);
+    }
+    return group;
+  }
+
+  private getAllureStep(step: TestStep, allureTest: AllureTest): AllureStep {
+    let allureStep = this.allureStepCache.get(step);
+    if (!allureStep) {
+      const parent = step.parent ? this.getAllureStep(step.parent, allureTest) : allureTest;
+      allureStep = parent.startStep(step.title);
```
just-boris, line 170: Same comment about `get*` name

just-boris (approved):
The implementation looks good, but there are a few stylistic suggestions

packages/allure-playwright/src/index.ts
dtinth: Oops, my bad 🤪

dtinth:
@just-boris Thank you for your comment! My original intention was to allow Playwright tests to invoke Allure APIs directly to add attachments and labels. I got stuck where when I realized that Playwright runs tests and reporters in different Node process.

I found that another PR is already solving my problem:

- https://github.com/allure-framework/allure-js/pull/408

IMO, I would prefer having that feature in first. Maybe I can rebase my changes on top of that after it got merged (it seems like our changes are not compatible with each other yet).

vovsemenv:
i love it 💚

baev:
> I found that another PR is already solving my problem

@dtinth  we decided to do the opposite, as #408 is still waiting for playwright 1.18 release and your one is good to go

baev:
@dtinth could you please rebase your PR?

skhomuti:
@dtinth are you still interested in this pull request?

dtinth:
Yes — sorry, I got caught up in other things 😞, will update it. Thanks for the kind reminder!

dtinth:
Rebased & review comments have been addressed.

baev:
thanks! 🎉

PR #417, opened 2022-01-27
Playwright - Add description attribute

unickq:
``` js
allure.description("description");
```

update for https://github.com/allure-framework/allure-js/pull/408#issuecomment-979772959

#### Checklist
- [x] [Sign Allure CLA][cla]

### DIFF
--- a/packages/allure-playwright/src/helpers.ts
+++ b/packages/allure-playwright/src/helpers.ts
@@ -1,16 +1,16 @@
 import test from "@playwright/test";
 import { Label, LabelName, Link, LinkType } from "allure-js-commons";
 
-
 export const ALLURE_METADATA_CONTENT_TYPE = "application/vnd.allure.metadata+json";
 export interface Metadata {
   labels?: Label[];
   links?: Link[];
+  description?: { value: string; html: boolean };
 }
 
 export class allure {
   static addMetadataAttachment(metadata: Metadata) {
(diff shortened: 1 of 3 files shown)


packages/allure-playwright/src/helpers.ts
```diff
@@ -22,6 +22,12 @@ export class allure {
     });
   }
 
+  static description(value: string, html = false) {
```
baev, line 25: I don't like the suggested method signature. Maybe use `description(value: string, opts?: { html: boolean})` instead? Or even introduce separate `descriptionHtml` method. 

Btw settings regular (non-html) description will only work if reset `descriptionHtml`. Otherwise `descriptionHtml` will be rendered no matter what value is set to `description`.

PS I rather deprecate html descriptions due to possible security issues

baev (changes requested):
could you please add a test as well?

packages/allure-playwright/src/helpers.ts
unickq: > PS I rather deprecate html descriptions due to possible security issues
Agree.

Then I just remove HTML from options at all

unickq: done @baev

baev (approved):

unickq:
@eroshenkoam can you have a look please?

Issue #405, opened 2021-11-12
[QUESTION] changing the "allure-report" folder location?


DvDream:
Hi, I'm using allure reporter with playwright, I can change the location of the "allure-results" folder. However, it seems like the "allure-report" folder is always saved in the main project directory and not in the folder I'd like to (the allure-result one). Is there a way to set it as well? Thank you!

PR #418, opened 2022-02-03
feat(playwright): add custom outputFolder in config

mxschmitt:
Fixes https://github.com/allure-framework/allure-js/issues/405

### DIFF
--- a/packages/allure-playwright/README.md
+++ b/packages/allure-playwright/README.md
@@ -45,6 +45,14 @@ set ALLURE_RESULTS_DIR=my-allure-results
 npx playwright test --reporter=line,allure-playwright
 ```
 
+or inside the config via:
+
+```js
+const config = {
+  reporter: [ ['allure-playwright', { outputFolder: 'my-allure-results' }] ],
+};
+```
+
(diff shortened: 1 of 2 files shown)


packages/allure-playwright/README.md
```diff
@@ -45,6 +45,14 @@ set ALLURE_RESULTS_DIR=my-allure-results
 npx playwright test --reporter=line,allure-playwright
 ```
 
+or inside the config via:
+
+```js
+const config = {
+  reporter: [ ['allure-playwright', { outputFolder: 'my-allure-results' }] ],
```
baev, line 52: maybe align the name with other modules? As far as I know, `resultsDir` is used in there.

baev (approved):
LGTM

packages/allure-playwright/README.md
mxschmitt: Inside the `@playwright/test` world `outputDir` and `outputFile` seems to be the norm: https://playwright.dev/docs/test-reporters#html-reporter, are you fine with aligning it that way?

PR #435, opened 2022-04-29
feat(playwright): implement environment allure function

vitalics:
### Context
Add allure helper function `environment`. 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-playwright/src/index.ts
```diff
@@ -168,6 +168,13 @@ class AllureReporter implements Reporter {
       );
     }
     allureTest.endTest();
+    const atLeastAllureAttachment = result.attachments.find(a => a.name === "environment.json");
+    if (atLeastAllureAttachment) {
+      const envInfo = atLeastAllureAttachment.body!.toString("utf8");
+      runtime.writeEnvironmentInfo(JSON.parse(envInfo));
+    } else {
+      runtime.writeEnvironmentInfo(process.env as Record<string, string>);
```
baev, line 176: this is unsafe and potentially could expose some secrets

baev (changes requested):

packages/allure-playwright/src/index.ts
vitalics: fixed now

packages/allure-playwright/src/index.ts
```diff
@@ -168,6 +168,13 @@ class AllureReporter implements Reporter {
       );
     }
     allureTest.endTest();
+    const atLeastAllureAttachment = result.attachments.find(a => a.name === "environment.json");
+    if (atLeastAllureAttachment) {
+      const envInfo = atLeastAllureAttachment.body!.toString("utf8");
+      runtime.writeEnvironmentInfo(JSON.parse(envInfo));
+    } else {
+      runtime.writeEnvironmentInfo(undefined);
```
baev, line 176: why write empty?

baev (changes requested):

packages/allure-playwright/src/index.ts
vitalics: because if no env variables - allure will use process.env by default
https://github.com/allure-framework/allure-js/blob/master/packages/allure-js-commons/src/AllureRuntime.ts#L59

vitalics: the essential thing here - is attaching environment info for run

baev: I don't see why we need unconditionally store all the process.env for each and every test result. I rather just remove `else` branch completely

packages/allure-playwright/src/index.ts
```diff
@@ -168,6 +168,13 @@ class AllureReporter implements Reporter {
       );
     }
     allureTest.endTest();
+    const atLeastAllureAttachment = result.attachments.find(a => a.name === "environment.json");
+    if (atLeastAllureAttachment) {
+      const envInfo = atLeastAllureAttachment.body!.toString("utf8");
```
baev, line 173: please process `undefined` body as well

packages/allure-playwright/src/helpers.ts
```diff
@@ -9,6 +9,16 @@ export interface Metadata {
 }
 
 export class allure {
+ /** 
+  * add environment info for test
+  * @default process.env
+  */
+  static environment(environment: Record<string, string | undefined> = process.env) {
+    test.info().attach("environment.json", {
```
vovsemenv, line 17: You don't need separate file for enviorment info. Instead you can just use `addMetadataAttachment` and add new metadata filed to it. It will be much easier to parse.

vovsemenv:
@vitalics please also look my other comment and we good to go

packages/allure-playwright/src/helpers.ts
vitalics: not agree with you, since allure generates `environment.json` as a separate file, I assume that is a good way to understand which environments are coming.

PR #448, opened 2022-06-01
chore(playwright): new reporter options for Allure TestOps compatibility

serembon:
### Context
When using the allure-playwright package, and uploading tests to Allure TestOps, the following inconveniences occur:

1. Reports contain Playwright`s API calls in substeps, and it hinders updating tests in TestOps.
api

2. By default, the suite label contains the path to the test file, and if the `allure.suite()` is declared, then the tests are duplicated, and `allure.suite()` is not parsed correctly.
suit

### Solution

Add options that can disable substeps with Playwright`s API calls  and not use the default value of the suite.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [ ]  Provide unit tests

### Unit tests
If unit tests are needed, then I need help with the fixtures


packages/allure-playwright/README.md
```diff
@@ -53,17 +53,31 @@ set ALLURE_RESULTS_DIR=my-allure-results
...
-or inside the config via:
+## Reporter options
+
+Some reporter settings can set by following options:
+
+| Option       | Description                                         | Default            |
+|--------------|-----------------------------------------------------|--------------------|
+| detail       | Hide `pw:api` and `hooks` steps in report           | `false`            |
+| outputFolder | Path to results folder                              | `./allure-results` |
+| suiteTitle   | Use test tittle instead of test file path in report | `false`            |
```
baev, line 64: typo

baev: BTW the description is unclear

baev (changes requested):
please revert all the formatting changes

packages/allure-playwright/README.md
serembon: Done

baev (approved):
Looks good!

baev:
BTW I rather introduce bunch of new labels instead of changing the default one. Allure TestOps provides an options to change the grouping and use any other label (mapped to custom field). But it's much harder to change the behaviour for Allure Report users

PR #459, opened 2022-06-27
Add host and thread labels for cucumberjs (fixes #454)

### LINKED ISSUE #454: CucumberJS: Add host & thread labels

lamartire:
### Context

fixes #454 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
+++ b/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
@@ -263,7 +263,7 @@ export class CucumberJSAllureFormatter extends Formatter {
 
     links.forEach((link) => currentTest?.addLink(link.url, link.name, link.type));
 
-    links.forEach((link) => this.currentTest?.addLink(link.url, link.name, link.type));
+    links.forEach((link) => currentTest?.addLink(link.url, link.name, link.type));
 
     // writting data tables as csv attachments
     pickle.steps.forEach((ps) => {
--- a/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
+++ b/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
@@ -267,8 +267,6 @@ export class CucumberJSAllureFormatter extends Formatter {
 
     links.forEach((link) => currentTest?.addLink(link.url, link.name, link.type));
 
-    links.forEach((link) => currentTest?.addLink(link.url, link.name, link.type));
-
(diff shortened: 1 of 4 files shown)


packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -178,6 +176,8 @@ export class CucumberJSAllureFormatter extends Formatter {
     this.currentTest = new AllureTest(this.allureRuntime, Date.now());
     this.currentTest.name = pickle.name;
 
+    this.currentTest?.addLabel(LabelName.HOST, os.hostname());
```
baev, line 179: use `ALLURE_HOST_NAME` provess.env if specified

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -178,6 +176,8 @@ export class CucumberJSAllureFormatter extends Formatter {
...
+    this.currentTest?.addLabel(LabelName.THREAD, process.getuid().toString());
```
baev, line 180: use ALLURE_THREAD_NAME process.env if specified

baev (changes requested):
lets add a possibility to set host/thread via env variables. Don't forget to add readme section about that as well

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
lamartire: 👍

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -178,6 +177,8 @@ export class CucumberJSAllureFormatter extends Formatter {
     this.currentTest = new AllureTest(this.allureRuntime, Date.now());
     this.currentTest.name = pickle.name;
 
+    this.currentTest?.addLabel(LabelName.HOST, ALLURE_HOST_NAME || os.hostname());
```
baev, line 180: if dns lookup is performed during `os.hostname()` call we may need to cache hostname into some static variable. Otherwise we will have some significant performance degradation as it's called once per scenario

lamartire: 👍

baev (approved):

PR #463, opened 2022-06-28
Provide `fullName` for test results

epszaw:
### Context
Adds feature `uri` (file path) as `fullName` parameter to be able to filter specs according [the documentation about filtering](https://github.com/cucumber/cucumber-js/blob/main/docs/filtering.md#names).

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev (changes requested):
as we use uri as full name it could be the same for multiple test cases we need to supply `testCaseId` field as well with some unique (and the same across different runs) scenario-specific hash (same testCaseId means the same test case created in ATO).

vovsemenv:
closed by https://github.com/allure-framework/allure-js/pull/530

PR #470, opened 2022-07-27
Add auto-detect of attachments extension  [playwright]

elipskiy:
The Playwright Reporter missed attachments whose extensions could not detect from content type. 

Add auto detect file extensions from the file name. 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-js-commons/src/AllureRuntime.ts
```diff
@@ -48,9 +48,21 @@ export class AllureRuntime {
   }
 
   writeAttachmentFromPath(
-    fromPath: PathLike,
+    fromPath: string,
```
baev, line 51: breaking change, please revert

baev (changes requested):

packages/allure-js-commons/src/AllureRuntime.ts
elipskiy: fixed

vovsemenv:
@elipskiy some tests are failed. Could you fix it?

vovsemenv (approved):

vovsemenv:
@baev we ready to go

elipskiy:
@baev please, merge this pr

elipskiy:
@baev @viclovsky

baev:
The PR isn't merged yet due to the following concerns:

1. Attachment file extension isn't really necessary as it will be detected automatically during report generation if correct content type is supplied. 
2. Besides that, we can use file content to detect correct content type (by using magic bytes) -- it's already implemented in both Allure Report & Allure TestOps
3. It makes sense to reuse existing file extension though. But I'm not sure about extname -- there are plenty of complex file extensions like `tar.gz` , and it may be confusing for some users (or not)
4. Why playwright? The main change is located in commons, so you need to add test to commons module.

baev:
fixed via https://github.com/allure-framework/allure-js/pull/956

PR #471, opened 2022-07-28
Remove yarn

lamartire:
### Context
Removes `yarn` with all the following artifacts

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


.github/workflows/build.yml
```diff
@@ -25,24 +25,34 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
-          cache: 'yarn'
 
-      - name: yarn install
+      - name: Restore npm cache
+        uses: actions/cache@v2
+        with:
+          path: |
```
mxschmitt, line 32: I think it's discouraged to cache node_modules and cache .npm instead see here: https://github.com/actions/cache/blob/main/examples.md#node---npm

lamartire: Good catch! Thank you for the comment 👍

lamartire:
Hey guys. Build is red because I haven't fixed eslint issues (I think, they have low priority), but build and test steps are green – that's enough to ensure about safe merge here.

vovsemenv (changes requested):
Why we need to add eslint deps in every package??
I created demo how it works without this changes https://github.com/vovsemenv/allure-js/pull/1.
Everything works, deps look cleaner and build become green.

packages/allure-mocha/package.json
```diff
@@ -29,10 +29,18 @@
     "@types/glob": "^7.1.4",
     "@types/mocha": "^9.0.0",
     "@types/node": "^16.7.8",
+    "@typescript-eslint/eslint-plugin": "^5.31.0",
+    "@typescript-eslint/parser": "^5.31.0",
+    "allure-commandline": "^2.18.1",
```
vovsemenv, line 34: We don't need `allure-commandline` devDependency in every package.
`npm run allure-report --workspace 'packages/allure-mocha'` works like a charm without it

lamartire: 👍

vovsemenv (approved):
Looks good

lamartire:
Better to left yarn due the publishing workflow

PR #484, opened 2022-08-20
Adding a check in onHookEnd to only call error function if it exists and is a function (fixes #482)

### LINKED ISSUE #482: Uncaught TypeError: hook.error is not a function MochaAllureReporter
**Describe the bug**
Getting an error when parallel mode is on while using Mocha

**To Reproduce**
1. Set parallel mode to true in mocha.rc config or use "--parallel" flag
2. Run the tests
3. Tests fail due to the following error: Uncaught TypeError: hook.error is not a function

**Expected behavior**
The tests should run successfully in parallel mode with Mocha allure reporter
**Screenshots**
image
image

**Additional context**
the following version packages are used:
    "allure-mocha": "^2.0.0-beta.19",
    "mocha": "^9.2.1"

jamesmortensen:
#### Checklist
- [X] [Sign Allure CLA][cla]
- [ ] Provide unit tests

Fixes #482 

The issue here is that, when running mocha in parallel mode with beforeEach or afterEach hooks, the onEndHook is called in MochaAllureReporter and the error() method is undefined.

Not sure about any unit tests for this.  The project unit tests weren't passing _before_ I made changes, so it's difficult to work with broken tests.  Instead, I have a reproduceable example here:   https://github.com/jamesmortensen/allure-js-reporter-issues#when-run-in-parallel-mode-with-hooks-hookerror-is-not-a-function

In the reproduceable example, run `npx mocha -p test-with-hooks.js` without the changes, to see the error occur, and run it with the changes to see the reporter function without errors and somewhat correctly.  (By somewhat correctly, there is this related issue https://github.com/allure-framework/allure-js/issues/485 where failed test cases in parallel mode are incorrectly reported as yellow/broken instead of red/failed.)

### DIFF
--- a/packages/allure-mocha/src/MochaAllureReporter.ts
+++ b/packages/allure-mocha/src/MochaAllureReporter.ts
@@ -81,6 +81,9 @@ export class MochaAllureReporter extends Mocha.reporters.Base {
   }
 
   private onHookEnd(hook: Mocha.Hook): void {
-    this.coreReporter.endHook(hook.error());
+    if(typeof(hook.error) === 'function')
+      this.coreReporter.endHook(hook.error());
+    else
+      this.coreReporter.endHook();
   }
 }


packages/allure-mocha/src/MochaAllureReporter.ts
```diff
@@ -81,6 +81,9 @@ export class MochaAllureReporter extends Mocha.reporters.Base {
   }
 
   private onHookEnd(hook: Mocha.Hook): void {
-    this.coreReporter.endHook(hook.error());
+    if(typeof(hook.error) === 'function')
+      this.coreReporter.endHook(hook.error());
```
lamartire, line 85: Could we set default value for `hook.error`? Like just stub function to avoid additional conditions?
Or we can use optional chaining operator to reach the same result:
```js
this.coreReporter.endHook(hook?.error?.());
```

lamartire (changes requested):

packages/allure-mocha/src/MochaAllureReporter.ts
jamesmortensen: @epszaw I will give it a try later in the month when I get some more time. Thanks for the feedback.

jamesmortensen: Hi @epszaw I just got back around to looking at this.  It looks like we can close this pull request.  I can see you added this change in this commit:  https://github.com/allure-framework/allure-js/commit/cbb34269c9ddcc6510e065693dfeb9b8c853beaf.  Thanks for adding it in.

delatrie:
Superseded by #510

PR #489, opened 2022-08-29
Move allure metadata to annotations

vovsemenv:
Move allure metadata to annotations.
This make allure metadata compatible with other reporters (i.e playwright-html reporter)

custom annotations docs https://playwright.dev/docs/test-annotations#custom-annotations


vovsemenv:
Not real benefits from this PR to allure itself

PR #494, opened 2022-09-06
fix cucumberjs reporter on windows

### LINKED ISSUE #492: Cannot run scenario after update to 2.0.0-beta.19
Hello there. I have been using beta 18 for running my tests and it runs fine. However when I update the package to beta 19 and start running the first scenario it's showing the error down below. Any idea what's going on?

node v16.17.0
npm 8.15.0

```
TypeError: process_1.default.getuid is not a function
    at CucumberJSAllureFormatter.onTestCaseStarted (..\node_modules\allure-cucumberjs\src\CucumberJSAllureReporter.ts:242:75)
    at CucumberJSAllureFormatter.parseEnvelope (..\node_modules\allure-cucumberjs\src\CucumberJSAllureReporter.ts:119:12)
    at EventEmitter.emit (node:events:532:35)
    at EventEmitter.emit (node:domain:475:12)
```

lamartire:
### Context
fixes #492

Use `process.pid` as fallback value for `THREAD` label

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
+++ b/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
@@ -232,6 +232,7 @@ export class CucumberJSAllureFormatter extends Formatter {
     const labels = this.parseTagsLabels(scenario?.tags || []);
     const links = this.parseTagsLinks(scenario?.tags || []);
     const currentTest = new AllureTest(this.allureRuntime, Date.now());
+    const thread = ALLURE_THREAD_NAME || process.pid.toString();
 
     this.runningTestsMap.set(data.id, currentTest);
     this.testCaseStartedMap.set(data.id, data);
@@ -239,9 +240,9 @@ export class CucumberJSAllureFormatter extends Formatter {
 
     currentTest.name = pickle.name;
     currentTest?.addLabel(LabelName.HOST, this.hostname);
-    currentTest?.addLabel(LabelName.THREAD, ALLURE_THREAD_NAME || process.getuid().toString());
     currentTest?.addLabel(LabelName.LANGUAGE, "javascript");
     currentTest?.addLabel(LabelName.FRAMEWORK, "cucumberjs");
+    currentTest?.addLabel(LabelName.THREAD, thread);
 
     if (doc?.feature) {
       currentTest.addLabel(LabelName.FEATURE, doc.feature.name);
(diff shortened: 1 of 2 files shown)


packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -232,17 +232,21 @@ export class CucumberJSAllureFormatter extends Formatter {
     const labels = this.parseTagsLabels(scenario?.tags || []);
     const links = this.parseTagsLinks(scenario?.tags || []);
     const currentTest = new AllureTest(this.allureRuntime, Date.now());
+    const thread = ALLURE_THREAD_NAME || process?.getuid()?.toString();
```
vovsemenv, line 235: maybe  we can use `process.pid` for compatibility

vovsemenv (changes requested):

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
lamartire: Why not, good catch

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -232,16 +232,17 @@ export class CucumberJSAllureFormatter extends Formatter {
     const labels = this.parseTagsLabels(scenario?.tags || []);
     const links = this.parseTagsLinks(scenario?.tags || []);
     const currentTest = new AllureTest(this.allureRuntime, Date.now());
+    const thread = ALLURE_THREAD_NAME || process?.getuid()?.toString() || process.pid.toString();
```
vovsemenv, line 235: I think we should completely switch to `procces.pid` and drop `getuid`

lamartire: `uid` and `pid` are different things, but yes, better to keep on thing here

vovsemenv: i think uid don't make sense in that case.

vovsemenv: when we wank to mark `process id` not `user id`

vovsemenv (approved):
Ok, it works.

But thread labels isn't right. Currently just mark all tests with current reporter process. Need worker Id from cucumber team i guess.

image

baev (approved):

PR #536, opened 2022-11-29
chore(deps): cleaup eslint deps from packages

vovsemenv:


epszaw (approved):
✅ The changes are reasonable. CI is green, so everything is fine.

PR #537, opened 2022-11-29
chore(deps): cleaup rimraf deps from packages

vovsemenv:


epszaw (approved):
✅ The changes are reasonable here

PR #538, opened 2022-11-29
chore(deps): cleanup prettier deps from packages

vovsemenv:


epszaw (approved):
✅ The change are reasonable here

PR #540, opened 2022-12-02
chore(license): delete licenses from sub-packages

vovsemenv:
There is no need to have license for each package


baev (changes requested):
JFYI playwright license has different copyright

baev (approved):

PR #543, opened 2022-12-09
feature(karma): add karma reporter

vovsemenv:


just-boris:
That project existed as an experiment to prove Allure architecture portability. There are no real use cases for such reporting in Karma, because it is strictly a unit-testing tool, with different use-cases.

I would vote for deprecating that package

vovsemenv:
Hi @just-boris thank you for the feedback. Actually i have the same thoughts about it did this only because the allure-karma package still receive  2000+ weekly downloads and stupid simple to implement.

baev:
as for today, karma is deprecated

PR #544, opened 2022-12-12
fix(build): fix build error

vovsemenv:


.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   build:
     name: "Build"
-    runs-on: ubuntu-latest
```
baev, line 16: According to https://github.com/actions/runner-images , `ubuntu-latest` stands for `ubuntu-20.04`. So why change?

baev (changes requested):

baev (approved):

.github/workflows/build.yml
vovsemenv: raised PR for runners https://github.com/actions/runner-images/pull/6766

PR #546, opened 2022-12-13
fix(cucumberjs): grouping test retries and attach datatables for steps

egerix:
### Context
When I switched to allure in cucumber v8, I encountered a number of problems that were not repeated in cucumber v6:
1) I can't install my CustomWorld, although the [documentation](https://github.com/allure-framework/allure-js/tree/master/packages/allure-cucumberjs#using-allure-api) has information about it. The reason is that the formatter explicitly sets CucumberAllureWorld after we specify our custom constructor
2) When using '--repeat', tests were duplicated instead of being placed in the history 
3) Attachments with the datatable are placed at the end of the test, not under the expected step. This is very unattractive if the test has several datatables. Previously, in Cucumber 6, they were placed under each step.

some pictures with changes:
**problem 2 before fix:**
image

**problem 2 after:**
image

**problem 3 before fix:**
image

**problem 3 after:**
image

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
+++ b/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
@@ -1,6 +1,6 @@
 import os from "os";
 import process from "process";
-import { Formatter } from "@cucumber/cucumber";
+import { Formatter, World } from "@cucumber/cucumber";
 import { IFormatterOptions } from "@cucumber/cucumber/lib/formatter";
 import TestCaseHookDefinition from "@cucumber/cucumber/lib/models/test_case_hook_definition";
 import * as messages from "@cucumber/messages";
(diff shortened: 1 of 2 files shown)


packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -293,7 +295,7 @@ export class CucumberJSAllureFormatter extends Formatter {
     const fullName = `${pickle.uri}#${pickle.name}`;
     currentTest.name = pickle.name;
     currentTest.fullName = fullName;
-    currentTest.testCaseId = md5(fullName);
```
baev, line 296: Please keep `testCaseId` as AllureTestOps uses it instead of `historyId`

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -146,9 +146,11 @@ export class CucumberJSAllureFormatter extends Formatter {
       }
       return message;
     };
-    // eslint-disable-next-line
-    // @ts-ignore
-    options.supportCodeLibrary.World = CucumberAllureWorld;
+    if (options.supportCodeLibrary.World === World) {
```
baev, line 149: please add test for that change

baev (changes requested):
Good job 👍

Could you please split the changes into 3 separate PR?

1. World change + test
2. Missing `historyId` + test (keep `testCaseId` and the test for it)
3. Data tables

egerix:
> Good job 👍
> 
> Could you please split the changes into 3 separate PR?
> 
> 1. World change + test
> 2. Missing `historyId` + test (keep `testCaseId` and the test for it)
> 3. Data tables

Okay, I'll close this pr and make three new ones

PR #547, opened 2022-12-15
fix(cucumberjs): grouping of retried tests

egerix:
### Context
split pr #546
When I switched to allure in cucumber v8, I encountered a number of problems that were not repeated in cucumber v6:
- When using '--repeat', tests were duplicated instead of being placed in the history 

some pictures with changes:
**problem before fix:**
image

**problem after fix:**
image

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


vovsemenv (approved):
Big thanks for the contribution !!!!
✅ Code looks good.

PR #553, opened 2022-12-15
chore(commons): delete mkdirp package

vovsemenv:
`fs.mkdirSync(path, {recursive:true})` doing the same thing without any packages


baev (changes requested):
add test

baev (approved):

PR #561, opened 2022-12-23
feature(commons): allure labels from env variables

vovsemenv:


packages/allure-js-commons/src/utils.ts
```diff
@@ -1,3 +1,21 @@
 import { createHash } from "crypto";
+import { env } from "process";
+import { AllureTest } from "./AllureTest";
 
 export const md5 = (data: string) => createHash("md5").update(data).digest("hex");
+
+export const attachLabelsFromEnv = (test: AllureTest) => {
```
baev, line 7: I would instead use a method that returns a list of labels from env

```ts
const providedLabelsFromEnv: () => AllureLabel[];
```

baev (changes requested):

baev (approved):

PR #576, opened 2023-01-19
feature(playwright): add layer helper

### LINKED ISSUE #575: Change test layers programmatically
We have a few types of tests and need to order the tests while they executed

maybe you can add to allure-playwright some function that can update the layer of the test?

for example:
  allure.layer('API')
  allure.layer('UI')

vovsemenv:
fixes #575

### DIFF
--- a/packages/allure-playwright/src/helpers.ts
+++ b/packages/allure-playwright/src/helpers.ts
@@ -79,7 +79,7 @@ export class allure {
 
   static layer(layerName: string) {
     this.label({
-      name: LabelName.PARENT_SUITE,
+      name: LabelName.LAYER,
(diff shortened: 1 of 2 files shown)


packages/allure-playwright/src/helpers.ts
```diff
@@ -78,6 +77,13 @@ export class allure {
     });
   }
 
+  static layer(layerName: string) {
+    this.label({
+      name: LabelName.PARENT_SUITE,
```
baev, line 82: should be "layer", not "parent_suite"

baev (changes requested):

baev (approved):

Issue #558, opened 2022-12-21
Error: No active group


vzhik54:
**Describe the bug**
I get this error after install dependencies(jest-jasmine2 allure-jasmine allure-js-commons @types/jasmine) when try to run any test from project.

**To Reproduce**
1. Install dependencies to project: npm i -D jest-jasmine2 allure-jasmine allure-js-commons @types/jasmine
2. Add allure-setup.ts and change jest.config.js file as per your instructions from readme file(Need to remove double line 'import { JasmineAllureReporter } from "allure-jasmine";' from example of readme file).
3. Run test: npm test( 'test' was added as value to "test" property of script section of packeage.json)

**Expected behavior**
There are no errors upon run test. Test runs successfully. Allure reports are saved to allure-results directory

**Desktop:**
 - OS: Windows 10 10.0.19042,
 - @types/jasmine: 4.3.1,
 - allure-jasmine: 2.0.0-beta.21,
 - allure-js-commons: 2.0.0-beta.21,
 - jest: 28.1.3,
 - jest-jasmine2: 29.3.1,

**Additional context**
Full error below:
`Error: No active group

      at JasmineAllureReporter.get currentGroup [as currentGroup] (node_modules/allure-jasmine/src/JasmineAllureReporter.ts:66:13)
      at node_modules/allure-jasmine/src/JasmineAllureReporter.ts:291:76
      at Object.<anonymous> (node_modules/allure-jasmine/src/JasmineAllureReporter.ts:254:40)`

vovsemenv:
Hello, thank you for participation. Readme types will be fixed soon.

About the main issue: I created stackblitz demo where i doing everything as you say and everything fine. Could you please change my demo to reproduce this problem?
https://stackblitz.com/edit/node-jf7ajr?file=example.test.js

vzhik54:
Thank you for your answer.
I investigated this issue more thoroughly and was able to reproduce it with the example I created. The error appears when using react + typescript.
I uploaded project to https://github.com/vzhik54/react-ts-allure-project

vovsemenv:
@vzhik54 thank you for the good reproduction found that this happens when we trying to use react-test-utils

reporoduction:
```ts
import "@testing-library/jest-dom";
import { render } from "@testing-library/react";

describe("ff", () => {
  test("gg", () => {
    //@ts-expect-error
    render(null);
    expect(1).toBe(1);
  });
});
```

vovsemenv:
will be fixed soon i guess

[...]

PR #579, opened 2023-01-23
fix(jasmine): fix global afterAll, beforeAll usages and support `@testing-library/react`

vovsemenv:
fix https://github.com/allure-framework/allure-js/issues/558


vovsemenv:
fix #558

baev (changes requested):
please add tests

baev (approved):

PR #583, opened 2023-01-25
fix(playwright): fix timeline threads for playwright

### LINKED ISSUE #555: Playwright timeline show too much threads
more info https://github.com/microsoft/playwright/issues/19519

vovsemenv:
fix #555


vovsemenv:
Hmmm we need to somehow set version of playwright for the users because this will not work for pw@<30

baev:
We can do it a safe way by checking that `parallelIndex` is in `result` before accessing it

baev (approved):

PR #607, opened 2023-02-14
feature(playwright): new imagediff

### LINKED ISSUE #416: Multiple screen diff sections
I'd like to compare screenshots of multiple areas in the page within the same test.

At the moment only it should be only one diff per test otherwise screen diff section is not shown

image

I'd like to see multiple collapsible screen diff entries ideally with the screenshot name

image

**Describe alternatives you've considered**
any other options to see multiple screen diffs per test

vovsemenv:
fixes #416


packages/allure-playwright/src/index.ts
```diff
@@ -201,12 +207,35 @@ class AllureReporter implements Reporter {
...
+        attachment.name.endsWith("-actual.png") ||
+        attachment.name.endsWith("-diff.png")
+      ) {
+        const diffEndRegexp = /-((expected)|(diff)|(actual))\.png/;
+
+        const pathWithoutEnd = attachment.path!.replace(diffEndRegexp, "");
+        if (this.processedDiffs.includes(pathWithoutEnd)) {
+          continue;
+        }
+        const actualBuffer = fs.readFileSync(`${pathWithoutEnd}-actual.png`).toString("base64");
```
baev, line 221: What if the file doesn't exist for some reason?

baev (changes requested):

packages/allure-playwright/test/attachment.spec.ts
```diff
@@ -67,16 +67,18 @@ test("should add snapshots correctly and provide a screenshot diff", async ({
     },
     (writer) => {
       return writer.tests[0].attachments.map((a) => {
```
epszaw, line 69: Just return attachments here, `map` is not needed here

packages/allure-playwright/README.md
```diff
@@ -323,3 +323,19 @@ export default {
...
+
+```ts
+import { test, expect } from "@playwright/test";
+test("screendiff", async ({ page }) => {
+  await page.goto("https://playwright.dev/");
+  await expect(page).toHaveScreenshot();
+});
+```
+
+if the screens doesn't match, the report will show the difference between the two screenshots.
```
epszaw, line 339: `If screenshots don't match, the report shows difference between them.`

packages/allure-playwright/src/index.ts
```diff
@@ -201,12 +207,48 @@ class AllureReporter implements Reporter {
         fileName = runtime.writeAttachmentFromPath(attachment.path!, attachment.contentType);
       }
 
-      if (attachment.name.endsWith("-expected.png")) {
-        allureTest.addAttachment("expected", attachment.contentType, fileName);
-      } else if (attachment.name.endsWith("-actual.png")) {
-        allureTest.addAttachment("actual", attachment.contentType, fileName);
-      } else if (attachment.name.endsWith("-diff.png")) {
-        allureTest.addAttachment("diff", attachment.contentType, fileName);
+      if (
```
epszaw, line 210: Use `diffEndRegexp` here

packages/allure-playwright/src/index.ts
```diff
@@ -201,12 +207,48 @@ class AllureReporter implements Reporter {
...
+        } catch (e) {}
+
+        try {
+          diffBase64 = fs.readFileSync(`${pathWithoutEnd}-diff.png`).toString("base64");
+        } catch (e) {}
+
+        const diffName = attachment.name.replace(diffEndRegexp, "");
+        const res = this.allureRuntime?.writeAttachment(
+          JSON.stringify({
+            expected: expectedBase64 && `data:image;base64,${expectedBase64}`,
```
epszaw, line 241: It would be nice, to move the logic to factory-function, what will generate expected string.
By the way, you can read the file by path in the function: `readImageAsBase64(path) -> string | undefined`

epszaw: And IMO better to use promisified versions of `fs` utilities

epszaw (approved):
✅ The code looks nice, don't have any question about it. Good job here! 👍

PR #620, opened 2023-02-20
add error notification when user try to use mocha reporter in paralle…

epszaw:
### Context
Add runtime error to prevent using the mocha reporter in parallel mode

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


vovsemenv (changes requested):
is it possible to add tests to check thar behavior?

epszaw:
> is it possible to add tests to check thar behavior?

I think, no, cuz we don't have access to spy anything in parallel thread

vovsemenv (approved):
Everything seems to be fine

PR #624, opened 2023-02-25
Add missing description methods for command steps

epszaw:
Adds description and description html methods to command allure step

### DIFF
--- a/packages/allure-js-commons/src/AllureCommandStep.ts
+++ b/packages/allure-js-commons/src/AllureCommandStep.ts
@@ -57,6 +57,10 @@ export interface AllureCommandStep<T = MetadataMessage> {
     content: Buffer | string,
     options: ContentType | string,
   ): void | Promise<void>;
+
+  description(content: string): void | Promise<void>;
+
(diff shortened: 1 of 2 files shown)


packages/allure-js-commons/src/AllureCommandStep.ts
```diff
@@ -57,6 +57,10 @@ export interface AllureCommandStep<T = MetadataMessage> {
     content: Buffer | string,
     options: ContentType | string,
   ): void | Promise<void>;
+
+  description(content: string): void | Promise<void>;
+
+  descriptionHtml(content: string): void | Promise<void>;
```
baev, line 63: `descriptionHtml` is very technical and has limitations. For regular users, we suggest always using `description` instead. Maybe only expose `description` for now?

epszaw: Ok, no problem

epszaw: 👍

PR #648, opened 2023-03-27
Added support for scoped tags

guiyomh:
### Context

It is not possible to make scoped tags. comme `env:dev` or `env:prod`.
This PR corrects the regex to support scoped tags.

Before when I have `@allure.label.tag=env:prod`, that doesn't works, because the name of the label was `tag=env` and the value was `prod`.

With this PR, now, the label is `tag` and the value is `env:prod`

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev (approved):

baev:
@guiyomh thanks!

PR #657, opened 2023-04-10
Pass original error from lambda-steps to the original test

### LINKED ISSUE #642: Allure-hermione step method doesn't hand over error outside
**Describe the bug**
We use the custom step method and divide the test into separate steps, if the test fails inside the step, then the step method catches the error, but doesn't hand over error outside.

**To Reproduce**
1. Use method "step" inside in any hermione test
2. Get error inside test intentionally
3. Сreate allure report
4. Look into report

**Expected behavior**
All test is failed on step that was failed

**Actual behavior**
Test is passed, we can see the error into step

**Screenshots**

epszaw:
### Context
fixes #642

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-js-commons/src/AllureCommandStep.ts
```diff
@@ -228,7 +250,7 @@ export class AllureCommandStepExecutable implements AllureCommandStep {
             start: startDate,
             stop: new Date().getTime(),
             stage: Stage.FINISHED,
-            status: Status.PASSED,
+            status: actualStatus,
```
vovsemenv, line 253: what about soft assertions? did hermione or cucumber have them? how it will work?

epszaw: What the soft assertions?

vovsemenv (approved):

packages/allure-js-commons/src/AllureCommandStep.ts
epszaw: Honestly I don't know about this feature in the frameworks. Anyway, if test will be marked as failed by any reason, result will be failed too.

vovsemenv (changes requested):
blocked for now

packages/allure-hermione/test/spec/steps.test.ts
```diff
@@ -3,27 +3,47 @@ import Hermione from "hermione";
...
+
+  describe("failed steps", () => {
+    beforeEach(async () => {
+      await hermione.run(["./test/fixtures/failedSteps.js"], {});
+    });
+
+    it("fails the test with original step error", () => {
+      const { status, statusDetails, steps } = hermione.allure.writer.results[0];
+
+      expect(status).eq(Status.FAILED);
```
vovsemenv, line 44: You also need to check steps for failed tests. Currently reporting of reporting of steps for failed steps if broken.

vovsemenv (changes requested):
Broken steps for failed tests need to be fixed

baev (approved):

Issue #663, opened 2023-04-20
Don't count repeatEach as a retry


alyaothman14:
**Describe the bug**
We are currently using repeatEach as a measure of the testcase performance. In allure, it counts the repeatEach as a retry and would mark the testcase as passed if one of the repeatEach fails which is incorrect.
With repeatEach, each testcase should be processed separately similar to how playwright report is 
![image](https://user-images.githubusercontent.com/87079479/233395496-7f753cd9-01a3-48a7-b1cc-c091f2db0350.png)
![image](https://user-images.githubusercontent.com/87079479/233395738-8c4b69c6-920a-40b3-af31-bfc661ffdc4c.png)

**Expected behavior**
Allure reports to behave the same as the playwright where is a separation between repeatEach and retry

baev:
already fixed

PR #664, opened 2023-04-20
fix:count in repeatEach so that testcases are distinctive as repeatEach is not the same as retry

alyaothman14:
### Context
Issue link: https://github.com/allure-framework/allure-js/issues/663
image

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-playwright/src/index.ts
+++ b/packages/allure-playwright/src/index.ts
@@ -102,6 +102,7 @@ class AllureReporter implements Reporter {
       allureTest.addLabel(LabelName.SUB_SUITE, suiteTitles.join(" > "));
     }
     const project = suite.project()!;
+    const repeatEach = project === null || project === void 0 ? void 0 : project.repeatEach;
     if (project?.name) {
       allureTest.addParameter("Project", project.name);
     }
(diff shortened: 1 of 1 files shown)


alyaothman14:
@baev I would really appreciate if this can be reviewed and released as the results in AllureTestOps is skewed because of this e.g the pipeline will fail but in testops it marks it as passed

packages/allure-playwright/src/index.ts
```diff
@@ -112,7 +113,7 @@ class AllureReporter implements Reporter {
       .join("/");
 
     const nameSuites = suiteTitles.length > 0 ? `${suiteTitles.join(" ")} ` : "";
-    const fullName = `${relativeFile}#${nameSuites}${test.title}`;
+    const fullName = `${relativeFile}#${nameSuites}${test.title}${repeatEach ? ` (repeat #${test.repeatEachIndex + 1})` : ""}`;
```
baev, line 116: Add it as a test parameter instead

baev:
```js
if (project?.repeatEach) {
  allureTest.addParameter("Repeation", test.repeatEachIndex + 1);
}
```

baev (changes requested):
could you please add a test for the change

packages/allure-playwright/src/index.ts
alyaothman14: Thanks updated

baev:
change included to https://github.com/allure-framework/allure-js/pull/676 

Thanks for the contribution!

PR #672, opened 2023-05-02
Add util to classify test errors

### LINKED ISSUE #667: Add utility to detect assertion errors
We need to detect when tests have been broken and when just failed. In Java implementation we have `AssertionException` class and it always means, that status should be `FAILED` in rest cases status should be equal to `BROKEN`.

In javascript we have many assertion libraries which potentially have different implementations, so we need to handle major of them (chai, node assert, etc.) automatically and provide to end-users ability to match their own assertion libraries through reporters.

1. When assertion error happened
2. Mark test as FAILED
---
1. When another kind of error happened
2. Mark test as BROKEN

epszaw:
### Context

fixes #667

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/package.json
+++ b/packages/allure-js-commons/package.json
@@ -42,6 +42,7 @@
     "@types/uuid": "^8.3.0",
     "allure-mocha": "workspace:packages/allure-mocha",
     "chai": "^4.3.4",
+    "expect": "^29.5.0",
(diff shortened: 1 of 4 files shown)


baev (approved):

baev:
@epszaw thanks!

PR #692, opened 2023-05-29
add currentTest init earlier than testStarted

### LINKED ISSUE #691: description method doesn't work for codeceptjs before hook, and event.started
i found that on my point of view we create `current.test` too late . right now it’s inside this method
```
testStarted(test) {
    this.currentTest = test;
    const allureTest = this.allureTestByCodeceptTest(test);
    const { labels } = (0, helpers_1.extractMeta)(test);
    if (allureTest) {
        labels.forEach((label) => {
            allureTest.addLabel(label.name, label.value);
        });
    }
}
```
and if i try add description in `before hook` or in
`event.dispatcher.on(event.test.started, async (test) => {`
in my repo - it will doesn’t add description because testStarted method works after my before.hook and event.test.started
that's why
```
this.description = (description) => {
    if (this.currentAllureTest) {
        this.currentAllureTest.description = description;
    }
};
```
doesn’t add description - we have no this.currentTest in the moment
but!! when we start suite we exactly know test!! by `suite.ctx.currentTest`
and if we init currentTest in 
```
suiteStarted(suite) {
        suite.tests.forEach((test) => {
            this.createTest(test);
        });
        this.currentTest = suite.ctx.currentTest
    }
```
description method will work in before hook and in event.test.started

hatufacci:
### Context

fixes #691 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-codeceptjs/src/index.ts
```diff
@@ -156,8 +158,6 @@ class AllureReporter {
   }
 
   testStarted(test: CodeceptTest & { tags: string[] }) {
-    this.currentTest = test;
```
baev, line 159: Maybe we should keep it? Just in case something went wrong like no `suite.ctx.currentTest` is set?

baev (changes requested):

packages/allure-codeceptjs/src/index.ts
hatufacci: @baev agree,
rolled back this.currentTest = test;

baev:
As far as I understand, you're trying to set a description to test from before hooks. But what if the suite has multiple tests? The description will be only set to the first one, is it right?

hatufacci:
@baev build was failed  because of another tests

hatufacci:
@vovsemenv could you take a look?

hatufacci:
@vovsemenv Hey, could you take a look? i've prepared it more than 2 month ago

packages/allure-codeceptjs/src/index.ts
```diff
@@ -219,6 +221,8 @@ class AllureReporter {
     this.currentTest = null;
     // @ts-ignore
     if (test.state === "failed" && test._retries !== 0 && currentRetry !== test._retries) {
+      // @ts-ignore
+      test.parent.ctx.currentTest = test;
```
vovsemenv, line 225: `Allure reporter` used purely to reporting info. But in this place you modify data. It can cause problems in the future then you using other plugins.

vovsemenv (changes requested):

packages/allure-codeceptjs/src/index.ts
hatufacci: @vovsemenv without this change in Report test data in retries missed, and we have only first attempt try

hatufacci: **with change** take a look on **Description** part:
 main run:
Screenshot 2023-08-03 at 14 03 12
attempt1
Screenshot 2023-08-03 at 14 03 47
attempt2:
Screenshot 2023-08-03 at 14 12 24

**without change**:
main run:
Screenshot 2023-08-03 at 14 09 41

 attempt1
Screenshot 2023-08-03 at 14 09 22

attempt2
Screenshot 2023-08-03 at 14 08 45

PR #696, opened 2023-06-05
Run lint job separately

epszaw:
### Context

It would be better to split test and lint job because sometimes it's difficult to understand which one has been failed, or one failed job can cancel another one and we need to repeat them many times after any fix.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


vovsemenv (approved):
looks fine

baev (approved):

PR #754, opened 2023-08-18
Add better description handler for cucumber features

### LINKED ISSUE #740: [Allure & Cucumber-JS] There is no description in the Allure Report
**Describe the bug**
All lines in feature files that come after Feature, Scenario or Scenario Outline are not displayed in the Allure report.

allure-cucumberjs: 2.4.0
@cucumber/cucumber: 9.3.0

**To Reproduce**
1. Prepare a feature file with approximately the following content
```
Feature: Some feature
  Feature's description
  
  Scenario: Some scenario
    Description of scenario

    Given some give step
    When some when step
    Then some than step
```
2. Implement the steps 
3. Run the test
4. Generate the report

**Expected behavior**
The "Description" block is expected, by analogy with how it is done in Java
image

**Actual behavior**
There is no "Description" block
image

epszaw:
### Context

fixes #740 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
+++ b/packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
@@ -300,9 +300,7 @@ export class CucumberJSAllureFormatter extends Formatter {
     const thread = data.workerId || ALLURE_THREAD_NAME || process.pid.toString();
     const fullName = `${pickle.uri}#${pickle.name}`;
     // sometimes description starts with "Description:", but it's not actually required
-    const description = scenario?.description
-      ? scenario.description.replace(/\s+Description:\s?/, "")
-      : doc?.feature?.description;
+    const description = scenario?.description || doc?.feature?.description || "";
     const testCaseId = md5(fullName);
 
     this.testCaseStartedMap.set(data.id, data);
@@ -310,7 +308,7 @@ export class CucumberJSAllureFormatter extends Formatter {
     this.currentTestsMap.set(data.id, currentTest);
 
     currentTest.name = pickle.name;
-    currentTest.description = description || "";
+    currentTest.description = description.trim();
     currentTest.fullName = fullName;
     currentTest.testCaseId = testCaseId;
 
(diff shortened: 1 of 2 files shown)


packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -299,13 +299,18 @@ export class CucumberJSAllureFormatter extends Formatter {
     const currentTest = new AllureTest(this.allureRuntime, Date.now());
     const thread = data.workerId || ALLURE_THREAD_NAME || process.pid.toString();
     const fullName = `${pickle.uri}#${pickle.name}`;
+    // sometimes description starts with "Description:", but it's not actually required
+    const description = scenario?.description
+      ? scenario.description.replace(/\s+Description:\s?/, "")
+      : doc?.feature?.description;
```
baev, line 305: Maybe always join the feature description with the scenario description?

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
```diff
@@ -299,13 +299,18 @@ export class CucumberJSAllureFormatter extends Formatter {
     const currentTest = new AllureTest(this.allureRuntime, Date.now());
     const thread = data.workerId || ALLURE_THREAD_NAME || process.pid.toString();
     const fullName = `${pickle.uri}#${pickle.name}`;
+    // sometimes description starts with "Description:", but it's not actually required
```
baev, line 302: Cucumber supports i18n. Is it always `Description:` or depends on the used language?

baev (changes requested):

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
epszaw: True, it doesn't relates on it at all, so we don't need to transform the description

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
epszaw: I afraid it'll produce strange descriptions with a lot of duplicated data. Guess, we should entirely replace feature description by a local one

packages/allure-cucumberjs/src/CucumberJSAllureReporter.ts
epszaw: Fixed and pushed

baev (approved):

PR #778, opened 2023-09-13
Fix wrapMethod return value

SergeyPirogov:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev:
@SergeyPirogov thanks! 

The issue is fixed in #776 and will be available in the next release

PR #845, opened 2024-01-18
Fix vitest files resolving

epszaw:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-vitest/src/index.ts
+++ b/packages/allure-vitest/src/index.ts
@@ -1,3 +1,4 @@
+/* eslint @typescript-eslint/require-await: 0 */
 import * as vitest from "vitest";
 import { LabelName, LinkType, MetadataMessage, ParameterOptions, Stage, Status, StepMetadata } from "allure-js-commons";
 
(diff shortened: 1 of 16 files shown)


baev (approved):

packages/allure-vitest/src/index.ts
```diff
@@ -27,82 +28,87 @@ const injectAllureMeta = (context: vitest.TaskContext) => {
   };
 };
 
-export const label = (context: vitest.TaskContext, name: string, value: string) => {
+export const label = async (context: vitest.TaskContext, name: string, value: string) => {
```
just-boris, line 31: Why is this needed? Do you have some plans adding async code here in the future? 

I can understand it for steps and attachments, but not much for labels.

There is a hidden issue that a missing await now may attach a label to an incorrect test

baev: We decided to provide async API for all the Allure methods for all the Allure integrations. The main reason is that we're currently working on a single Allure Facade, exposed by commons, that all the frameworks will implement. 

That will allow us to create second-level integrations, e.g. with tools like Axios. There is an issue with such integrations right now since it should know what AllureApi to use to add an attachment. 

So, the solution would be a single Facade, similar to how it's implemented for other languages: 

```js
const { allure } from "allure-js-commons";

it("it could be any test framework", async() => {
  await allure.step(...);
});
```

Having this in mind, should AllureFacade be async or not? It should, at least for the `step` and `attachment` API. For some frameworks (hi, playwright), the bus we use to communicate from the worker process to the main (reporter) process is asynchronous by design, so all the Allure methods are asynchronous.

We don't like to have this difference between different frameworks, nor do we want to support both async and non-async APIs. 

That's why we decided to have all the Allure API methods async.

PR #904, opened 2024-03-11
Generic crypto

epszaw:
#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests


noomorph:
I'm curious: Which JS+Allure integrations support browser (non-Node) environments? From your PR, it seems that you support such. 🤔

epszaw:
Hey, @noomorph! Presently, we have a new cypress integration which provides code both for browser and node, and we need to do [similar tricks](https://github.com/allure-framework/allure-js/blob/main/packages/allure-cypress/src/model.ts#L2) to make the code work. The PR is a small step to make Allure Commons platform-agnostic as a part of #897.

noomorph:
@epszaw, thank you. Yeah, now I see why you need that. The last time I remembered in-browser test runners being widely used was in the days of `jasmine` and `karma`. Every day you learn something. :)

PR #974, opened 2024-05-24
Ensure results dir created before write result

### LINKED ISSUE #894: Ensure the results directory exists when writing a file in AllureRuntime
AllureRuntime creates the results directory only once (in its constructor).
Some testing tools like Cypress support interactive testing with long-running sessions. If the directory gets deleted, each further attempt to write an Allure file will fail until the directory is recreated by hand or until the tool is restarted.

We should check if the directory exists before writing a file. If it's not, we should create it.

_Originally posted by @delatrie in https://github.com/allure-framework/allure-js/pull/815#discussion_r1504774392_

epszaw:
### Context

fixes #894

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/test/spec/FileSystemAllureWriter.spec.ts
+++ b/packages/allure-js-commons/test/spec/FileSystemAllureWriter.spec.ts
@@ -39,15 +39,18 @@ describe("FileSystemAllureWriter", () => {
     expect(actualContent.toString("utf8")).toBe(data);
   });
 
-  it("Should create allure-report nested path", () => {
+  it("creates allure-report nested path every time writer write something", () => {
     const tmpReportPath = path.join(os.tmpdir(), `./allure-testing-dir/${randomUUID()}`);
     const config: Config = {
       writer: new FileSystemAllureWriter({
         resultsDir: tmpReportPath,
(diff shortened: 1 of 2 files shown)


packages/allure-js-commons/test/spec/FileSystemAllureWriter.spec.ts
```diff
@@ -39,15 +39,18 @@ describe("FileSystemAllureWriter", () => {
     expect(actualContent.toString("utf8")).toBe(data);
   });
 
-  it("Should create allure-report nested path", () => {
+  it("creates allure-report nested path every time writer write something", () => {
```
delatrie, line 42: Given the spec, a proper test would be to delete the directory after `writeTest` and check if it's re-created after another one.

delatrie (approved):

packages/allure-js-commons/test/spec/FileSystemAllureWriter.spec.ts
epszaw: 👍

baev (approved):

PR #991, opened 2024-06-04
Commons improvements

baev:
Commons refactoring mainly related to attachments.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-playwright/src/legacy.ts
```diff
@@ -44,105 +44,105 @@ export const allurePlaywrightLegacyApi: AllurePlaywrightLegacyApi = {
   /**
    * @deprecated please use import { label } from "allure-js-commons" instead.
    */
-  label: (...args) => allure.label(...args),
+  label: (...args) => Promise.resolve(allure.label(...args)),
```
epszaw, line 47: We don't need to wrap the methods into `Promise.resolve` because the wrapped methods return a promise or undefined, so we can add `await` keyword for both versions and nothing will be broken.

packages/allure-js-commons/src/sdk/reporter/types.ts
```diff
@@ -86,9 +86,9 @@ export interface Writer {
 
   writeGroup(result: TestResultContainer): void;
 
-  writeAttachment(distFileName: string, content: Buffer | string, encoding?: BufferEncoding): void;
+  writeAttachment(distFileName: string, content: Buffer): void;
 
-  writeAttachmentFromPath(from: string, distFileName: string): void;
+  writeAttachmentFromPath(distFileName: string, from: string): void;
```
epszaw, line 91: `to` will be much easier to understand what the argument does

packages/allure-playwright/src/legacy.ts
baev: `Promise.resolve` is required to convert `PromiseLike` to `Promise`

packages/allure-js-commons/src/sdk/reporter/types.ts
baev: It's just aligned with the `writeAttachment` method. We can revisit the naming in separate PR

delatrie (approved):

PR #995, opened 2024-06-06
Fix Allure Mocha ESM parallel; add deprecation warnings to the legacy API

delatrie:
### Context

Some fixes for Allure Mocha:

  - Fix the ESM version in parallel mode: now `import.meta` is used instead of `__dirname`. The `babel-plugin-transform-import-meta` plugin is used to transpile that construct to CommonJS.
  - deprecation warnings were added to the legacy API
  - MochaAllureReporter was renamed to AllureMochaReporter
  - `runtime.ts` was removed (it's not in use now)

Test improvements for Allure Mocha:

  - switching between `esm` and `cjs` samples (controlled by `ALLURE_MOCHA_TEST_SPEC_FORMAT`, the default is `esm`).
  - switching between the CLI and code runners, where the code runner can be in the `esm` or `cjs` format (controlled by `ALLURE_MOCHA_TEST_RUNNER`, the default is `cli`).

Linting fixes:

  - `dist` folders were removed from ESLint globally
  - no-require rules were disabled for Allure Mocha CommonJS files

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-mocha/test/fixtures/runner.js
```diff
@@ -0,0 +1,55 @@
...
+// cjs: const dirname = __dirname;
+// esm: const dirname = path.dirname(fileURLToPath(import.meta.url));
+
+let emitFiles = false;
+let parallel = false;
+const requires = [];
+
+const sepIndex = process.argv.indexOf("--");
+const args = sepIndex === -1 ? [] : process.argv.splice(sepIndex);
+for (let i = 0; i < args.length; i++) {
```
epszaw, line 20: Use `forEach` instead

packages/allure-mocha/test/spec/api/runtime/legacy/parameters.test.ts
```diff
@@ -48,7 +48,7 @@ describe("legacy runtime parameters API", () => {
         parameters: [
           {
             name: "foo",
-            // eslint-disable-next-line
+            // eslint-disable-next-line @typescript-eslint/quotes
             value: '{"key":10}',
```
epszaw, line 52: Just FYI, you can use `String.raw`{"key":10}` to bypass the linter and formatter

epszaw (approved):

packages/allure-mocha/test/fixtures/runner.js
delatrie: Can't use forEach: there is a top-level await in the ESM version. But I've changed the loop to `for (const arg of args) ...`.

Issue #1079, opened 2024-07-26
Allure Cypress 3.0.0-beta: hooks are reported inconsistently


delatrie:
**Describe the bug**
The following fixture reporting issues should be fixed in Allure Cypress:

  - Before/after fixtures are reported for all tests in the run.
  - Before each/after each fixtures are reported for siblings of the their target.

**To Reproduce**
1. Create a spec file and define a test in the file scope.
1. In the spec file, define a suite (`describe`) with two tests inside.
3. Define four hooks (`before`, `after`, `beforeEach`, and `afterEach`) inside the suite.
4. Execute `cypress run`.
5. Check `before` and `after` fixtures are listed for the file scoped test.
6. Check `beforeEach` and `afterEach` fixtures are listed twice for both tests inside the suite.

**Expected behavior**
1. `before` and `after` hooks should be reported for tests in the suite only.
2. `beforeEach` and `afterEach` hooks should be reported once per each test in the suite

**Desktop (please complete the following information):**
 - Version 3.0.0-beta.1

Issue #1078, opened 2024-07-26
Allure Cypress 3.0.0-beta lose test results if the run gets interrupted.


delatrie:
**Describe the bug**
Cypress test are reported at the end of the run. That makes it possible to lose test results for tests that are in fact completed. A typical scenario is a CI pipeline, which may interrupt the action after a timeout.

**To Reproduce**
1. Create a Cypress project with two spec files, each containing a test. Install `allure-cypress@3.0.0-beta.7`.
2. Execute `cypress run`.
3. Once Cypress starts executing the second spec, press `CTRL+C`.
4. No Allure files are generated.

**Expected behavior**
The result file for the test in the first spec file should be generated.

**Desktop (please complete the following information):**
 - Version 3.0.0-beta.7

**Additional context**
Ideally, test results should be reported in real time. That might get tricky with Cypress, so we most probably should resort to the second best option: reporting test results after each spec.

Issue #1072, opened 2024-07-25
allure-cypress test failures in before & beforeEach blocks are not reported


iamrragu:
**Describe the bug**
If the cypress tests fail in the `before` or `beforeEach` block, they are not being reported in the allure-results.

**To Reproduce**
```
describe('Test Suite', { tags: '@test' }, () => {
  before(() => {
    expect(false).to.eq(true);
  });

  it('should pass', () => {
    expect(true).to.eq(true);
  });
});
```

When the test passes we get 2 files under allure-results

```
{
  "uuid": "de42dc8f-5fc6-4b23-84eb-2c49b0024fca",
  "name": "\"before all\" hook",
  "children": [
    "8496a4a2-bfff-4334-b446-febbbb83b73c"
  ],
  "befores": [
    {
      "status": "passed",
      "statusDetails": {},
      "stage": "finished",
      "steps": [],
      "attachments": [],
      "parameters": [],
      "start": 1721890224434,
      "name": "\"before all\" hook",
      "stop": 1721890224653
    }
  ],
  "afters": []
}
```
```
{
  "uuid": "8496a4a2-bfff-4334-b446-febbbb83b73c",
  "name": "should pass",
  "historyId": "7deea96eaba6ccacf320023fe1e01d97:d41d8cd98f00b204e9800998ecf8427e",
  "status": "passed",
  "statusDetails": {},
  "stage": "finished",
  "steps": [],
  "attachments": [],
  "parameters": [],
  "labels": [
    {
      "name": "language",
      "value": "javascript"
    },
    {
      "name": "framework",
      "value": "cypress"
    },
    {
      "name": "parentSuite",
      "value": "Test Suite"
    }
  ],
  "links": [],
  "start": 1721890224325,
  "fullName": "cypress/e2e/test.cy.ts#Test Suite should pass",
  "testCaseId": "7deea96eaba6ccacf320023fe1e01d97",
  "stop": 1721890224651
}
```
and when the test fails in `before` or `beforEach` block the `allure-results` folder is empty

**Expected behavior**
Test failures in test hooks should be available in allure-results and reports.

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] - `Mac OS Sonoma 14.5`
 - Browser [e.g. chrome, safari] - `Chrome`, `Electron`
 - Version - `cypress@13.13.0` and `allure-cypress@3.0.0-beta.7`

**Additional context**
[...]

Issue #1070, opened 2024-07-23
allure-cypress : getting error  - 'Converting circular structure to JSON'


delatrie:
### Discussed in https://github.com/orgs/allure-framework/discussions/2650

<div type='discussions-op-text'>

<sup>Originally posted by **pranit-lqc** July 22, 2024</sup>
We are using Vue 3 component on the UI side.

Version,
"allure-commandline": "^2.29.0",
"allure-cypress": "^3.0.0-beta.7",
"cypress": "13.5.1",

Getting below error:
<img src="https://github.com/user-attachments/assets/fe16722c-3199-46a2-abdc-b1cf95bf0556" width="20%">

without the Allure plugin, we don't see this error.
</div>

delatrie:
Presumably, we're trying to stringify too much here:

https://github.com/allure-framework/allure-js/blob/e4e77f74c2c47751085b43f4afa27c1d90ae7769/packages/allure-cypress/src/index.ts#L542

Karczewski:
Hi, I'm encountering the same issue. 
Instead of using `JSON.stringify`, you can use something like this:
```
function safeStringify(obj, replacer, spaces) {
  try {
    return JSON.stringify(obj, replacer, spaces);
  } catch (error) {
    return `[Unserializable: ${error.message}]`;
  }
};
```

perrsikk:
Hello 
Issue reproduces with latest allure-cypress version  "3.0.0-beta.10"
![image](https://github.com/user-attachments/assets/5964ac0d-4d85-4eb6-a58d-b91f2bd23086)

baev:
@perrsikk thanks for the info

BlessedLongsword:
I am also having this issue. For now, it has happened when doing a `cy.wrap()` inside a `.each()` loop of a list of DOM elements, or at the end of a  `.within()` block.

For example:

- `within()` block 
```js
    cy.get('[status="positive"]').first().within(() => {
      cy.contains('Created');
      cy.get([icon="clear"]).click();
    });
```

- `each()` block
```js
cy.get('[data-testid="loc-list"]').find('list')
            .should('have.length', data.length)
            .each(($loc, index) => {
              const locName = data.at(index).name;
              cy.wrap($loc).should('contain.text', locName);
            });
```

![image](https://github.com/user-attachments/assets/b973d841-5d57-4798-a4a7-557bdaab916d)

BlessedLongsword:
[...]

PR #1109, opened 2024-08-09
feat(allure-cypress): new real-time lifecycle, hooks, and test plan

delatrie:
### Context
The PR contains the following major changes of Allure Cypress:

#### Real-time reporting
Previously, Allure Cypress reported the results in a batched fashion when the test run ended. The new implementation employs:

  - an updated set of messages
  - message queueing on the browser side
  - queue flushing to the Node side in the plugin's `afterEach` hook
  - the tail flushing in the plugin's `after` hook

The messages are flushed using Cypress'es tasks mechanism. It works reliably only in hooks, that's why `index.ts` defines `afterEach` and `after` hooks.

In interactive mode, the spec reporting is finalized in Mocha's `run end` event. In `cypress run` mode the finalization is done in Cypress'es `after:spec` event.

`environment.properties` and `categories.json` are reported in `after:run`.

#### New hooks implementation

The new implementation supports all types of hooks. Spec-level `after` hooks are supported by intercepting the `after` calls and injecting the plugin's `after` hooks right next to user-defined ones. That allows us to issue the reporting chronologically after the user-defined `after` is completed and its corresponding messages are queued.

#### New test plan implementation

The previous implementation requested the test plan from the browser side through the Cypress task. It applies the test plan in a `beforeEach` hook and skipped tests with `this.skip()`.

The downside is that, while correctly skipping the tests, this approach fails to disable the associated hooks, meaning all the `before` and `after` hooks will be executed. Given those hooks often contain tasks like spinning up test servers, initializing databases, etc, that may slow down the execution substantially.

The new implementation uses Cypress environment variables to pass the test plan down the specs. The only requirement is to pass the Cypress configuration to `allureCypress`:

```javascript
import { defineConfig } from "cypress";
import { allureCypress } from "allure-cypress/reporter";

export default defineConfig({
  e2e: {
    setupNodeEvents: (on, config) => {
      allureCypress(on, config, { /* Allure options goes here */ });
      return config;
    },
    // ...
  },
});
```

> [!NOTE]
> For backward compatibility, passing Allure configuration as the 2nd argument also works.

fixes https://github.com/allure-framework/allure-js/issues/1078
fixes https://github.com/allure-framework/allure-js/issues/1079
fixes https://github.com/allure-framework/allure-js/issues/1072
fixes https://github.com/allure-framework/allure-js/issues/1070

### Other changes

  - New README for Allure Cypress

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


baev (approved):

baev:
fixes https://github.com/allure-framework/allure-js/issues/1079

PR #1111, opened 2024-08-12
Run tests on windows

baev:
#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests


delatrie:
### Typescript configuration files for testing
Previously, test files were covered by `tsconfig.test.json` files in the root of each package. That didn't play well with some IDEs because, when analyzing a file, they expect `tsconfig.json` to be present in the file directory or its parent. In this PR I've put those config files in the `test` directories and renamed them to `tsconfig.json`. Vitest and ESlint configs have been adjusted.

### Casing of environment labels
Given the following two environment variables:

  - `ALLURE_LABEL_foo=bar`
  - `ALLURE_LABEL_FOO=baz`

When we give them to a process, it receives either two distinct variables (Linux, Mac) or just one (Windows).
Since the behavior differs across OS, I've removed the corresponding checks in the environment variable tests.



### OS- and CWD-consistency of test results

All packages now calculate `fullName`, `package`, `testCaseId`, and `historyId` in a way that doesn't depend on OS and CWD:

  - On Windows, paths are converted to POSIX-paths (`\` chars are replaced with `/`).
  - Relative paths are resolved against the project root (the nearest directory that contains `package.json`). If there is no project root, the CWD is used as a fallback.

The list of packages, fixed by this PR:

  - allure-cucumberjs
  - allure-cypress
  - allure-jasmine
  - allure-jest
  - allure-mocha
  - allure-playwright
  - allure-vitest

Allure-codeceptjs uses allure-mocha under the hood, which means it automatically inherits those behaviors.

Newman-reporter-allure doesn't use FS paths, so it doesn't need to be fixed.

### Allure-mocha testCaseId collision

Allure-mocha mistakenly doesn't use the test file path when calculating `testCaseId` of a test. This PR fixes that.

epszaw (approved):

PR #1127, opened 2024-08-30
feat(allure-cypress): impose some limits on step parameter values created from Cypress command arguments

### LINKED ISSUE #1070: allure-cypress : getting error  - 'Converting circular structure to JSON'

delatrie:
### Context

Allure Cypress converts Cypress commands to Allure steps. Command argument values are converted to Allure step parameter values:

  - strings are left unmodified,
  - values of other types are converted with `JSON.stringify(value, null, 2)`.

The drawbacks:

  - A `TypeError: Converting circular structure to JSON` is thrown if the value is an object with circular references. This is often the case when using Cypress commands with complex UI objects like Vue components, etc. See some examples in issue #1070.
  - Complex objects and long strings/arrays can easily blow the report to ridiculous sizes

### `serialize` enhancements

We have an existing function `serialize` defined in `allure-js-commons/sdk/reporter/utils`, which behaves similarly to what Allure Cypress does with Cypress step arguments. This PR:

1. Moves it to `allure-js-commons/sdk/utils` to make it usable from environments other than Node.js.
2. Adds an optional argument, which, if given, must be an object with two optional properties: `maxDepth`, `maxLength`. The value of each property must be an integer number greater than or equal to 0 (which is the default). The value of zero means the property doesn't affect `serialize`.
3. Implements the following constraints imposed on the result of `serialize`:
      - If `maxDepth` is given and is greater than 0, properties that create nesting levels beyond that limit are precluded from serialization.
      - If `maxLength` is given and is greater than 0, and if the length of the resulting string is greater than that value, the remaining part of the string is replaced with `"..."`.

#### Examples

```javascript
const obj = {
  foo: {
    bar: {
      baz: ["Lorem", "Ipsum"],
      qux: "Dolor",
    },
  },
  bar: {},
};
const obj1 = {};
obj1.ref = obj1;
serialize(obj1); // returns "{}"

const obj2 = {
  // this is nesting level 1
  foo: {
    // this is nesting level 2
    bar: {
      // this is nesting level 3
      /* ... */ 
    },
    baz: "Lorem ipsum",
  },
};
serialize(obj2, { maxDepth: 2 }); // returns '{"foo":{"bar":"Lorem ipsum"}}'

serialize({ foo: "Lorem ipsum" }, { maxLength: 10 }); // returns '{"foo":"Lo...'
```

allure-cypress now uses `serialize` to convert Cypress command arguments to strings.

### Configuration changes

The PR introduces two new configuration properties `maxArgumentLength` and `maxArgumentDepth` to customize the abovementioned limits for allure-cypress. The properties are grouped under the `stepsFromCommands` option:

```javascript
import { defineConfig } from "cypress";
import { allureCypress } from "allure-cypress/reporter";

export default defineConfig({
  e2e: {
    setupNodeEvents: (on, config) => {
      allureCypress(on, config, {
        stepsFromCommands: {
          maxArgumentLength: 64,
          maxArgumentDepth: 5,
        },
      });

      return config;
    },
    // other Cypress options
  },
});
```

The default values are 128 (for `maxArgumentLength`) and 3 (for `maxArgumentDepth`).

Fixes #1070

### Other minor changes

  - improve maps and sets support by `serialize`: now they may nest in other objects and contain circular references, which are properly removed upon serialization. The values are serialized as arrays instead of the `"[object Map]"` and `"[object Set]"` literals.
  - document why the test plan feature may not work (because of the missing 2nd argument of `allureCypress`)

### Potential future changes

  - We may use the new `serialize` implementation to take objects of any types in `attachment` (currently, only strings and buffers are allowed);
  - May include an extra flag (e.g., `disabled`) into the `stepsFromCommands` object of the Allure Cypress config to preclude Cypress steps from getting into Allure Report (see allure-framework/allure2#2712).

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cypress/src/model.ts
+++ b/packages/allure-cypress/src/model.ts
@@ -8,10 +8,7 @@ export const ALLURE_REPORT_STEP_COMMAND = "__allure_report_step_command__";
 
 export type AllureCypressConfig = ReporterConfig & {
   videoOnFailOnly?: boolean;
-  stepsFromCommands?: {
-    maxArgumentLength?: number;
-    maxArgumentDepth?: number;
-  };
+  stepsFromCommands?: Partial<AllureSpecState["config"]["stepsFromCommands"]>;
 };
 
 export type CypressSuite = Mocha.Suite & {
--- a/packages/allure-cypress/src/model.ts
+++ b/packages/allure-cypress/src/model.ts
@@ -8,6 +8,10 @@ export const ALLURE_REPORT_STEP_COMMAND = "__allure_report_step_command__";
 
 export type AllureCypressConfig = ReporterConfig & {
   videoOnFailOnly?: boolean;
+  stepsFromCommands?: {
+    maxArgumentLength?: number;
+    maxArgumentDepth?: number;
+  };
 };
 
 export type CypressSuite = Mocha.Suite & {
@@ -174,6 +178,12 @@ export type SpecContext = {
 };
 
 export type AllureSpecState = {
+  config: {
+    stepsFromCommands: {
+      maxArgumentLength: number;
+      maxArgumentDepth: number;
+    };
+  };
   initialized: boolean;
   testPlan: TestPlanV1 | null | undefined;
   messages: CypressMessage[];
--- a/packages/allure-cypress/src/reporter.ts
+++ b/packages/allure-cypress/src/reporter.ts
@@ -32,7 +32,7 @@ import type {
   CypressTestStartMessage,
   SpecContext,
 } from "./model.js";
-import { last } from "./utils.js";
+import { defaultRuntimeConfig, last } from "./utils.js";
 
 export class AllureCypress {
   allureRuntime: ReporterRuntime;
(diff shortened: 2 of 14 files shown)


packages/allure-cypress/src/reporter.ts
```diff
@@ -489,19 +489,28 @@ export class AllureCypress {
...
 });
 
-/**
- * Explicitly enables the selective run feature.
- * @param config The Cypress configuration.
- */
-export const enableTestPlan = (config: Cypress.PluginConfigOptions) => {
-  config.env.allure = getInitialSpecState();
-  return config;
+const applyDefaultsToRuntimeConfig = ({
```
epszaw, line 499: The name of the function says that it returns the whole of the config, but modified one, but actually, it returns a single field. IMO `getConfigDefaults` would fit the logic much more, or let's change the logic to make the function return whole the config with applied defaults.

packages/allure-cypress/src/runtime.ts
```diff
@@ -361,11 +363,14 @@ export const reportTestSkip = (test: CypressTest) => {
 };
 
 export const reportCommandStart = (command: CypressCommand) => {
+  const { maxArgumentDepth, maxArgumentLength } = getConfig().stepsFromCommands;
```
epszaw, line 366: You also can do deep restructirization:

```js
const { stepsFromCommands: { maxArgumentDepth, maxArgumentLength } } = getConfig();
```

packages/allure-cypress/src/utils.ts
```diff
@@ -5,6 +5,13 @@ import { ALLURE_REPORT_STEP_COMMAND, ALLURE_REPORT_SYSTEM_HOOK } from "./model.j
 import type { CypressCommand, CypressHook, CypressSuite, CypressTest } from "./model.js";
 import { getAllureTestPlan } from "./state.js";
 
+export const defaultRuntimeConfig = {
```
epszaw, line 8: Let's name the constant using `UPPER_SNAKE_CASE`

epszaw (approved):
Good job here! I noticed nothing serious, so if you want to fix nothing – just skip the comments.

packages/allure-cypress/src/model.ts
```diff
@@ -174,6 +178,12 @@ export type SpecContext = {
 };
 
 export type AllureSpecState = {
+  config: {
```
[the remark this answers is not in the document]
delatrie: Make sense! But I've changed it in the other way around, because we need it to be fully defined here, but `Partial<..>` when exposed to users.

packages/allure-cypress/src/reporter.ts
delatrie: I've renamed it to `getRuntimeConfigDefaults` so as not to confuse it with the user-exposed AllureCypressConfig instance.

packages/allure-cypress/src/runtime.ts
delatrie: 100%!

packages/allure-cypress/src/utils.ts
delatrie: Agree!

PR #1205, opened 2024-12-03
feat(cypress): command log steps rework

### LINKED ISSUE #1059: [Cypress + Cucumber] Execution commands are not grouped into Given-When-Then steps
**Describe the bug**
Execution commands are not grouped into Given-When-Then steps

Project structure: Cypress (latest) + @badeball/cypress-cucumber-preprocessor (latest )+ allure-cypress (latest: 3.0.0-beta.6)

**To Reproduce**
1. Open attached project [cypress-cucumber-allure.zip](https://github.com/user-attachments/files/16192580/cypress-cucumber-allure.zip)
2. `npm install` to initialize 
3. run `cypress run`  to run all tests and produce allure-results
4. generate AllureReport and open test result

**Expected behavior**
Scenario commands are grouped as per Given-When-Then steps like it was in a previous Cypress-Allure adapter  (@shelex/cypress-allure-plugin).

**Screenshots**
image

**Additional context**
I do understand that I can wrap commands into allure.step() but this doesn't make any sense in case of Cucumber. GWT steps are already containers for the commands.

delatrie:
## Context

Allure Cypress automatically converts commands into Allure steps. It does so by listening `command:start` and `command:end`, 

The PR reworks this mechanism to be based on the command log instead. It aligns with our general attitude to represent things as they are reported by the framework itself: the command log is the main reporting mechanism that Cypress provides to users in interactive mode.

> [!NOTE]
> The Cypress command log supports events, e.g., the `fetch` events, which allow tracking HTTP requests made with the fetch API. Events like this are not the subject of this PR. The implementation currently doesn't create steps from them. We may support such events in the future.

### Child commands and assertions

Previously, Allure Cypress didn't include child commands, e.g., assertions. Now, it includes them as substeps of their parent command.

<details>
<summary>Example</summary>

```js
it("assertions", () => {
  cy.wrap(1 + 1)
    .should("eq", 2)
    .and((v) => {
      expect(v).lessThan(3);
      return v + 1;
    })
    .and("eq", 3);
});
```

<details>
<summary>The Comman log</summary>

image

</details>

<details>
<summary>Allure Report before</summary>

image

</details>

<details>
<summary>Allure Report after</summary>

image

</details>

</details>

### Properties of command log entries

Previously, Allure Cypress used to create step parameters from the `attributes.args` list of a command object. Now it uses the `attributes.consoleProps` function of a command log entry. This is the function Cypress uses to print the properties in the console when a user clicks on an entry in interactive mode.

The main difference is that `consoleProps` provides key-value pairs instead of a list of values, which allows as to use the actual names instead of syntetic ones. Additionally, many Cypress commands include extra useful data in such props.

<details>
<summary>Example</summary>

```js
it("properties", () => {
  cy.visit("/");
  cy.get("input[type=text]")
    .should("have.attr", "placeholder")
    .and("eq", "Search without being tracked");
});
```

<details>
<summary>The Comman log and the console</summary>

image

image

</details>

<details>
<summary>Allure Report before</summary>

image

</details>

<details>
<summary>Allure Report after</summary>

image

</details>

</details>

### Custom command log entries

Cypress allow users to add extra entries to the command log with `Cypress.log`. Allure Cypress now correctly converts them into steps. Child entries (`type: "child"`) are included as substeps of the nearest running parent step (if any).

<details>
<summary>Example</summary>

```js
it("custom log", () => {
  Cypress.log({ name: "start", message: "asserting", type: "parent" });
  cy.wrap(1 + 1).should((v) => {
    expect(v).eq(2);
    Cypress.log({ name: "done", message: "asserting", type: "child" });
  });
});
```
<details>
<summary>The Comman log</summary>

image

</details>

<details>
<summary>Allure Report</summary>

image

</details>

</details>

### Hidden commands and log entries
Some commands take the `{ log: true }` option. Cypress precludes such commands from the command log. Allure Cypress doesn't convert them to steps as well.

<details>
<summary>Example</summary>

```js
it("hidden logs", () => {
  cy.wrap(1 + 1, { log: false }).should("eq", 2);
  cy.wrap(1 + 2, { log: false }).should("eq", 3);
});
```

<details>
<summary>The Comman log</summary>

image

</details>

<details>
<summary>Allure Report</summary>

image

</details>

</details>

### Command log groups

Command log entries can be grouped by providing the `{ groupStart: true }` option to `Cypress.log`. Allure Cypress converts such entries to steps. Every step created until `log.endGroup` is called becomes a substep. This applies to steps from the command log as well as to steps created with `allure.step`.

<details>
<summary>Example</summary>

```js
it("log group", () => {
  const group = Cypress.log({
    name: "asserting",
    message: "the result of 1 + 1",
    type: "parent",
    groupStart: true,
  });
  cy.wrap(1 + 1).should("eq", 2).then(() => {
    group.endGroup();
  });
});
```

<details>
<summary>The Comman log</summary>

image

</details>

<details>
<summary>Allure Report</summary>

image

</details>

</details>

> [!NOTE]
> Since [@badeball/cypress-cucumber-preprocessor](https://www.npmjs.com/package/@badeball/cypress-cucumber-preprocessor) uses the `Cypress.log` API to group command log entries of a cucumber step, such grouping is automatically preserved by Allure Cypress. This wasn't the case earlier (see #1059).

### Screenshots

Allure Cypress now uses the [attachment step feature](https://allurereport.org/docs/steps/#attachment-only-steps) to report screenshots created with `cy.screenshot` or created by Cypress automatically when a test fails.

<details>
<summary>Example</summary>

```js
it("screenshot", () => {
  cy.visit("/");
  cy.get(".is-not-mobile-device")
    .screenshot()
    .find("input[type=text]")
    .screenshot("the input");
});
```

<details>
<summary>Allure Report before</summary>

image

</details>

<details>
<summary>Allure Report after</summary>

image

</details>

</details>

## Other changes

  - `serialize` (in `allure-js-commons/sdk/utils`) now supports a new `replacer` option. A value must be a function with the following signature `(this: unknown, key: string, value: unknown) => unknown`. If provided, it's used by `serialize` to pre-process object property values.
  - Allure Cypress now works better with other listeners of `Cypress.on("fail", ...)`, which are created by a user or a 3rd party plugin. The exception is throws only if there are no subsequent listeners. Otherwise, it passes the responsibility of throwing to the next listener.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

Closes #1059.


packages/allure-cypress/src/browser/patching.ts
```diff
@@ -0,0 +1,130 @@
...
+ * Patches the `after` function, to inject reporting of spec-level
+ * `after` hooks defined by the user.
+ */
+export const enableScopeLevelAfterHookReporting = () => {
+  const [getSuiteDepth, incSuiteDepth, decSuiteDepth] = createSuiteDepthCounterState();
+  patchDescribe(incSuiteDepth, decSuiteDepth);
+  patchAfter(getSuiteDepth);
+};
+
+const createSuiteDepthCounterState = (): [get: () => number, inc: () => void, dec: () => void] => {
```
epszaw, line 19: Guess, it would be much more convenient to work with object, not an array here because we need to keep in mind an order of the functions to use them properly, but in the case of an object, it's stupidly simple:

```js
const { count, increment, decrement } = createSuiteDepthCounterState()

// ...
```

packages/allure-cypress/src/browser/patching.ts
```diff
@@ -0,0 +1,130 @@
...
+    () => {
+      suiteDepth++;
+    },
+    () => {
+      suiteDepth--;
+    },
+  ];
+};
+
+const patchDescribe = (incSuiteDepth: () => void, decSuiteDepth: () => void) => {
```
epszaw, line 32: Here we can just pass the object I described above

packages/allure-cypress/src/browser/utils.ts
```diff
@@ -0,0 +1,140 @@
+import { Status } from "allure-js-commons";
+import type { StatusDetails } from "allure-js-commons";
+import { extractMetadataFromString, getMessageAndTraceFromError, getStatusFromError } from "allure-js-commons/sdk";
+import type { CypressHook, CypressStepStopMessage, CypressSuite, CypressTest, StepDescriptor } from "../types.js";
+import { ALLURE_REPORT_SYSTEM_HOOK } from "./events/mocha.js";
+import { getAllureState, getProjectDir } from "./state.js";
+import { resolveStepStatus } from "./steps.js";
+
+export const getFileNameFromPath = (path: string) => path.substring(path.lastIndexOf(IS_WIN ? "\\" : "/") + 1);
```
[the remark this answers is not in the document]
delatrie: Those are browser-friendly (and limited) versions of getPosixPath + node:path functions, which I had to write because those from allure-js-commons are Node-only. I don't think other packages need them.

Also, I don't think we should use `slash`. It does the same replacing except a special handling for unprocessed Windows paths (`\\?\...`), which doesn't apply in our case (I don't think Cypress supports uses such paths). We would still need the branching on the platform (Unix file names can contain backslashes), so it doesn't save us much overall.

packages/allure-cypress/src/browser/utils.ts
```diff
@@ -0,0 +1,140 @@
...
+    },
+  };
+};
+
+const getSkipReason = (hookTitle: string, suite: CypressSuite) => {
+  const suiteName = suite.title ? `'${suite.title}'` : "root";
+  return `'${hookTitle}' defined in the ${suiteName} suite has failed`;
+};
+
+const IS_WIN = Cypress.platform === "win32";
```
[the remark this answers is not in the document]
delatrie: Done

packages/allure-cypress/src/browser/patching.ts
delatrie: Nice suggestion, done

packages/allure-cypress/src/browser/patching.ts
delatrie: Done

PR #1221, opened 2025-01-15
Enchance global labels definition

epszaw:
### Context

Previously it was possible to use array of `Label` type to configure global labels. Now users can define them by `Record<string, string | string[]>` as well.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/src/sdk/reporter/types.ts
+++ b/packages/allure-js-commons/src/sdk/reporter/types.ts
@@ -47,10 +47,12 @@ export type LinkConfig<TOpts extends LinkTypeOptions = LinkTypeOptions> = Partia
 
 export type WriterDescriptor = [cls: string, ...args: readonly unknown[]] | string;
 
+export type GlobalLabelsConfig = Record<string, string | string[]>
+
 export interface ReporterConfig {
   readonly resultsDir?: string;
   readonly links?: LinkConfig;
-  readonly globalLabels?: Label[];
+  readonly globalLabels?: Label[] | GlobalLabelsConfig;
   readonly listeners?: LifecycleListener[];
   readonly environmentInfo?: EnvironmentInfo;
   readonly categories?: Category[];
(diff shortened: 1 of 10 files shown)


packages/allure-js-commons/src/sdk/reporter/types.ts
```diff
@@ -47,10 +47,12 @@ export type LinkConfig<TOpts extends LinkTypeOptions = LinkTypeOptions> = Partia
 
 export type WriterDescriptor = [cls: string, ...args: readonly unknown[]] | string;
 
+export type GlobalLabelsConfig = Record<string, string | string[]>;
```
delatrie, line 50: BTW, if we define this thing as `Partial<Record<LabelName, string | string[]>> & Record<string, string | string[]>`, the IDE will suggest well-known label names (epic, host, etc.) to a user while still allowing them to provide custom ones.

epszaw: Good point. Pushed

delatrie (approved):

packages/allure-js-commons/src/sdk/reporter/types.ts
delatrie: No, TypeScript erases the LabelName part, making the key type just `string`, which won't help with IDE suggestions.
You have to use type intersection to force TypeScript, keeping the LabelName part.

delatrie: You may check it yourself:

```ts
type GlobalLabelConfig = Record<LabelName | string, string | string[]>;

const v: GlobalLabelConfig = {
  /* no suggestions here */
};
```

```ts
type GlobalLabelConfig = Partial<Record<LabelName, string | string[]>> & Record<string, string | string[]>;

const v: GlobalLabelConfig = {
  /* IDE suggests well-known label names */
};
```

epszaw: It does work pretty well in idea:

image

But you're right, it doesn't work in LSP. So I enhanced the type

delatrie: I guess that it won't work in IDEs that use TypeScript itself to infer the types for suggestions, but it may work in IDEs that do some extra job to extract more info about the type.

baev (approved):

PR #1227, opened 2025-01-21
Support spaces in the title metadata format

epszaw:
#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/src/sdk/utils.ts
+++ b/packages/allure-js-commons/src/sdk/utils.ts
@@ -73,7 +73,7 @@ type AllureTitleMetadataMatch = RegExpMatchArray & {
   };
 };
 
-export const allureTitleMetadataRegexp = /@?allure\.(?<type>\S+)[:=]("[^"]+"|'[^']+'|`[^`]+`|\S+)/;
+export const allureTitleMetadataRegexp = /(?:^|\s)@?allure\.(?<type>\S+)[:=]("[^"]+"|'[^']+'|`[^`]+`|\S+)/;
 export const allureTitleMetadataRegexpGlobal = new RegExp(allureTitleMetadataRegexp, "g");
 export const allureIdRegexp = /(?:^|\s)@?allure\.id[:=](?<id>\S+)/;
 export const allureLabelRegexp = /(?:^|\s)@?allure\.label\.(?<name>[^:=\s]+)[:=](?<value>[^\s]+)/;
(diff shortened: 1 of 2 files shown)


packages/allure-js-commons/src/sdk/utils.ts
```diff
@@ -56,13 +63,31 @@ export const getMessageAndTraceFromError = (error: Error | { message?: string; s
...
+  groups: {
+    type?: string;
+    v1?: string;
+    v2?: string;
+    v3?: string;
+    v4?: string;
+  };
+};
+
+export const allureTitleMetadataRegexp = /@?allure\.(?<type>\S+)[:=]("[^"]+"|'[^']+'|`[^`]+`|\S+)/;
```
delatrie, line 76: We should bind this RE to avoid polynomial complexity in a way similar to how we did it for id and label REs (see [here](https://github.com/allure-framework/allure-js/pull/1065)).

This time we can't fully prevent backtracking, but it will be limited to three occurrences (one per each type of quote), which is OK.

```suggestion
export const allureTitleMetadataRegexp = /(?:^|\s)@?allure\.(?<type>\S+)[:=]("[^"]+"|'[^']+'|`[^`]+`|\S+)/;
```

delatrie (changes requested):

packages/allure-js-commons/src/sdk/utils.ts
epszaw: I think, it wouldn't be unreasonable 👍

packages/allure-js-commons/src/sdk/utils.ts
```diff
@@ -56,13 +63,31 @@ export const getMessageAndTraceFromError = (error: Error | { message?: string; s
...
+export const allureIdRegexp = /(?:^|\s)@?allure\.id[:=](?<id>\S+)/;
 export const allureLabelRegexp = /(?:^|\s)@?allure\.label\.(?<name>[^:=\s]+)[:=](?<value>[^\s]+)/;
-export const allureLabelRegexpGlobal = new RegExp(allureLabelRegexp, "g");
+
+export const getTypeFromAllureTitleMetadataMatch = (match: AllureTitleMetadataMatch) => {
+  return match?.[1];
+};
+
+export const getValueFromAllureTitleMetadataMatch = (match: AllureTitleMetadataMatch) => {
+  return match?.[2]?.replace?.(/['"`]/g, "");
```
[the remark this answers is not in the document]
epszaw: Another good point 👍

delatrie (approved):
Looks good to me, nice job! 👍

PR #1243, opened 2025-02-04
feat(codeceptjs): add support for meta steps

### LINKED ISSUE #1242: Sub-steps when using CodeceptJS page objects
CodeceptJS contains a feature called [page object](https://codecept.io/pageobjects/#pageobject) which allows you to group steps into methods instead of just writing a long test line-by-line, potentially making it difficult to read through. Until now I've been using [allure-legacy](https://github.com/codeceptjs/allure-legacy) plugin. Because it is not maintained any longer I'm looking to switch to new allure-codeceptjs instead.

One of the main differences I've noticed between the 2 is how steps within page objects are presented in the report. For additional context here is the small sample test I'm using. Scenario:
```javascript
Scenario('login test',  ({ I, login }) => {
    I.amOnPage('https://github.com')
    I.waitForVisible('.lp-IntroHero')
    login.openLoginPage()
})
```
Login page object: 
```javascript
const { I } = inject()

module.exports = {
    openLoginPage() {
        I.waitForVisible('.HeaderMenu-link--sign-in')
        I.click('.HeaderMenu-link--sign-in')
        I.waitForVisible('.auth-form-body')
    }
}
```
The legacy plugin reports these as sub-steps within the page object method which IMO makes it much easier to read:
Image

Newer plugin instead just reports everything as a top-level step so there's no indication which steps come from a page object and which are defined inside scenario itself:
Image

I think it would be nice to have the same nested sub-steps in the new plugin as well. Are there any plans to support this? Maybe at least have an option that can be enabled if such nesting is needed?

**Describe alternatives you've considered**
I've thought about somehow modifying this using a custom plugin on top of `allure-codeceptjs` plugin but since all of the logic for defining steps is contained within that plugin I don't think I can do this (other than re-writing the whole plugin myself).

**Additional context**
As far as I understand the code the legacy plugin achieved this using something called [metasteps](https://github.com/codeceptjs/allure-legacy/blob/8c31bd2d0736487edaa7d92a813562b2571cef49/index.js#L295). I've noticed there is a reference to metastep in this repository within [Typescript definitions](https://github.com/allure-framework/allure-js/blob/fb678d21f7652f21fa102b1fd219f57b3fa9bfda/packages/allure-codeceptjs/src/model.ts#L21) but it doesn't seem to be used anywhere?

For reference these are dependencies from `package.json` used in the above example. I'm using Playwright but it should be the same with other drivers.
``` 
"dependencies": {
    "@codeceptjs/allure-legacy": "1.0.2",
    "allure-codeceptjs": "3.1.0",
    "codeceptjs": "3.6.10",
    "playwright": "1.50.1"
}
```
Node version: v20.10.0

baev:
### Context

fixes #1242

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-codeceptjs/src/reporter.ts
+++ b/packages/allure-codeceptjs/src/reporter.ts
@@ -2,12 +2,20 @@ import { event, recorder } from "codeceptjs";
 import type * as Mocha from "mocha";
 import { env } from "node:process";
 import { LabelName, Stage, Status, type StepResult } from "allure-js-commons";
-import { getMessageAndTraceFromError, getStatusFromError, isMetadataTag } from "allure-js-commons/sdk";
+import { getMessageAndTraceFromError, getStatusFromError, isMetadataTag, stripAnsi } from "allure-js-commons/sdk";
 import AllureMochaReporter from "allure-mocha";
 import type { CodeceptBddStep, CodeceptError, CodeceptStep } from "./model.js";
 
+interface MetaStep {
+  name: string;
+  id: string;
+}
+
+const MAX_META_STEP_NESTING = 10;
+
 export class AllureCodeceptJsReporter extends AllureMochaReporter {
   protected currentBddStep?: string;
+  protected metaStepStack: MetaStep[] = [];
 
   constructor(runner: Mocha.Runner, opts: Mocha.MochaOptions, isInWorker: boolean) {
     super(runner, opts, isInWorker);
@@ -18,6 +26,7 @@ export class AllureCodeceptJsReporter extends AllureMochaReporter {
     // Test
     event.dispatcher.on(event.test.before, this.testStarted.bind(this));
     event.dispatcher.on(event.test.failed, this.testFailed.bind(this));
+    event.dispatcher.on(event.test.finished, this.testFinished.bind(this));
     // Step
     event.dispatcher.on(event.step.started, this.stepStarted.bind(this));
     event.dispatcher.on(event.step.passed, this.stepPassed.bind(this));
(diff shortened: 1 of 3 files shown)


packages/allure-codeceptjs/src/reporter.ts
```diff
@@ -47,24 +56,95 @@ export class AllureCodeceptJsReporter extends AllureMochaReporter {
...
+
+    let index = 0;
+    while (
+      index < Math.min(this.metaStepStack.length, stepPath.length) &&
+      this.metaStepStack[index].name === stepPath[index]
+    ) {
+      index++;
+    }
+
+    for (let i = index; i < this.metaStepStack.length; i++) {
```
epszaw, line 129: Why you don't use just `forEach`?

packages/allure-codeceptjs/src/reporter.ts
```diff
@@ -47,24 +56,95 @@ export class AllureCodeceptJsReporter extends AllureMochaReporter {
...
+      const id = this.metaStepStack[i].id;
+      this.runtime.updateStep(id, (result) => {
+        result.status = Status.PASSED;
+        result.stage = Stage.FINISHED;
+      });
+      this.runtime.stopStep(id);
+    }
+
+    for (let i = index; i < stepPath.length; i++) {
```
epszaw, line 138: Same here. Let's use just `forEach`

epszaw (approved):

packages/allure-codeceptjs/src/reporter.ts
baev: the start index isn't 0

packages/allure-codeceptjs/src/reporter.ts
baev: the start index isn't 0

PR #1274, opened 2025-03-18
feat: add function to attach Playwright Trace

todti:
### Context
Add special contentType `application/playwright-trace` for Playwright Trace archives.

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/src/facade.ts
+++ b/packages/allure-js-commons/src/facade.ts
@@ -77,7 +77,9 @@ export const attachment = (
 };
 
 export const attachTrace = (name: string, path: string) => {
-  return callRuntimeMethod("attachmentFromPath", name, path, { contentType: "application/playwright-trace" });
+  return callRuntimeMethod("attachmentFromPath", name, path, {
+    contentType: "application/vnd.allure.playwright-trace",
+  });
 };
 
 export const attachmentPath = (
--- a/packages/allure-js-commons/src/facade.ts
+++ b/packages/allure-js-commons/src/facade.ts
@@ -76,6 +76,10 @@ export const attachment = (
   return callRuntimeMethod("attachment", name, content, opts);
 };
 
+export const attachTrace = (name: string, path: string) => {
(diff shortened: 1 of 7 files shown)


packages/allure-js-commons/src/facade.ts
```diff
@@ -76,6 +76,12 @@ export const attachment = (
   return callRuntimeMethod("attachment", name, content, opts);
 };
 
+export const attachTrace = (name: string, path: string) => {
```
epszaw, line 79: It shouldn't be a part of the facade due to the trace is a playwright's only abstraction

packages/allure-playwright/src/runtime.ts
```diff
@@ -51,6 +51,10 @@ export class AllurePlaywrightTestRuntime extends MessageTestRuntime {
     await test.info().attach(name, { path, contentType: options.contentType });
   }
 
+  async attachTrace(name: string, content: Buffer) {
```
epszaw, line 54: This function already does what we need, so there isn't necessary to add something to the commons' facade

epszaw: Also, add a test for the function

epszaw (changes requested):

epszaw (approved):

packages/allure-playwright/src/index.ts
```diff
@@ -623,9 +623,11 @@ export class AllureReporter implements ReporterV2 {
     } else if (!existsSync(attachment.path!)) {
       return;
     } else {
-      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+      const contentType =
+        attachment.name === "trace" ? "application/vnd.allure.playwright-trace" : attachment.contentType;
```
epszaw, line 627: Check the attachment content type too to be sure because we can have `trace.jpg`, `trace.js` etc.:

`attachment.name === "trace" && attachment.contentType === "application/zip" ? ...`

epszaw (changes requested):

epszaw (approved):

PR #1348, opened 2025-09-05
pw attachments great again

### LINKED ISSUE #1334: PW attachments are added only to beforeAll hook from other steps
**Describe the bug**
When the method that attaches data to the step is used in the beforeAll hook, all attachments are being added only to it

**To Reproduce**
```
import { test } from "@playwright/test"

test.describe("Scratch", () => {
  test.beforeAll(async () => {
    await example("test")
  })
  test("test", async () => {
    await test.step("test2", async () => {
      await example("test2")
    })

    await test.step("test3", async () => {
      await example("test3")
    })
  })
})

const example = async (some: string) => {
  await test.info().attach("test", { body: some })
}
```

**Expected behavior**
All attachments should be added to the corresponding steps

**Screenshots**
Image

**Additional context**
"allure-commandline": "2.34.1"
"allure-playwright": "3.3.3"
"playwright": "1.54.1"

If the hook doesn't call in the stack that method, then attachments are added correctly
Image

Refs:
#1278 
#1303

todti:
### Context

Fix  #1334 

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-playwright/package.json
+++ b/packages/allure-playwright/package.json
@@ -93,6 +93,6 @@
     "vitest": "^3.2.4"
   },
   "peerDependencies": {
-    "@playwright/test": ">=1.36.0"
+    "@playwright/test": ">=1.53.0"
   }
 }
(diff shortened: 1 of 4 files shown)


romek-buildops:
🫶

PR #1350, opened 2025-09-10
Fix attachment assignment order

todti:
### Context

Attachments with identical names were being incorrectly assigned to test steps and hooks.
Tests cover scenarios with both identical and different attachment names across hooks and test steps.
Thanks to @simple-elf for finding the bug!

#### Checklist
- [ ] [Sign Allure CLA][cla]
- [ ] Provide unit tests


todti:
@mvolkert Hi! Could you please attach some examples here? I'll check what happened.

mvolkert:
@todti 
Im using "@playwright/test": "1.55.1"

In playwright.config.ts I got: `screenshot: {mode: 'only-on-failure', fullPage: true},`
With "allure-playwright": "3.4.0" or earlier it looked like this (the error screenshot was clearly visible):
allure_playwirght_before
With "allure-playwright": "3.4.1" the screenshot gets displayed somewhere beneath afterHooks:
allure_playwirght_after
Sometimes not at all (but it is taken by playwright).
It seams like the error screenshot gets liked to a step and sometimes it even gets lost during this linking

PR #1360, opened 2025-10-01
fix cypress origin serialization (fix #1306)

### LINKED ISSUE #1306: [allure-cypress] Unhandled rejection CypressError: `cy.origin()` could not serialize the subject due to one of its properties not being supported by the structured clone algorithm.
**Describe the bug**
Allure fails to handle some tricks with cy.origin()

I've created a project with a test that reproduce it. Unfortunately it's not possible to upload zipped files here and I also cannot upload a folder.
So, please look at the NOT zipped folder on the [Google Drive.](https://drive.google.com/drive/folders/1ufy3ZbXcofhnC_nn9W6JLlvHLNOmbQ9a?usp=sharing)

I've spent a week to catch it in my project and to share with you on a test project. Hope it helps a lot.

**To Reproduce**
Please add the following to your hosts file:
127.0.0.1 main-app.local
127.0.0.1 admin-portal.local
127.0.0.1 brand-portal.local
127.0.0.1 api.local

In package.json file, execute the next command to start local server:
"serve": "http-server ./test-site -p 80 --cors"

Then open Cypress ("test:crossdomain:open": "cypress open --config-file cypress.crossdomain.config.ts") 
 and run the test in file: allure-bug-try.e2e.cy.ts

Make sure you start test with clean session (it will not reproduce if the first session is restored rather than created)

**Actual behavior**
Test hangs forever

Image

**Expected behavior**
I don't know, but it should fail like that. Now it hangs forever.

**Additional context**
It started in v. 3.3.0, in version 3.2.0 my production tests where I caught that bug failed with another error, so it's hard to say if this issue was present or not.

todti:
- fixed #1306 when Cy.origin() just hang with error and blocked next steps

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-cypress/src/browser/serialize.ts
```diff
@@ -1,8 +1,43 @@
...
-export default (value: unknown) =>
-  isDomObject(value) ? stringifyAsDom(value) : serialize(value, getSerializeOptions());
+export default (value: unknown) => {
+  return isDomObject(value) ? stringifyAsDom(value) : serializeAsObject(value);
+};
+
+const isPlainObject = (value: unknown): value is Record<string, unknown> =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
+
+const stripAncestorRefs = (value: unknown, ancestors: object[] = []): unknown => {
```
delatrie: `serialize` already does this. Is there a case it can't handle?

PR #1400, opened 2026-02-02
fix(allure-js): fix test duplicating by adding package name to path

### LINKED ISSUE #1323: Tests from different packages might be treated as retries of each other
**Describe the bug**

Different tests get the same `fullName`, `testCaseId`, and `historyId` if:

  - They are from different packages (e.g., in a multirepo).
  - And they have the same file path and suite names.

Allure Report (both v2 and v3) will display only one such test, hiding the others behind the "Retries" tab.

**To Reproduce**

Since the issue affects the `allure-js` repo, the easiest way to reproduce it is as follows:

1. Clone this repo, checkout `tags/v.3.2.2`, install deps, and build the code.
2. `yarn workspaces foreach -Avvt run vitest run ./test/spec/globalLabels.test.ts -t "^should handle global labels$" --passWithNoTests`.
3. `allure serve $(find ./packages/allure-*/out/allure-results -type d)` (for Windows, `allure serve @(gi ./packages/allure-*/out/allure-results/)`).
4. See the error: only one test is displayed in the trees. Other tests are hidden as the test's retries, rather than being separate tests.

**Expected behavior**
Tests of different packages must be treated as separate tests in all cases.

**Screenshots**

Image

Image

Image

**Smartphone (please complete the following information):**

**Additional context**

We normalize fullNames relative to the package.json, which means we can read the package's name from it in most cases.

formaceft-93:
Fix test duplicating by adding package name from package.json to fileName 

Closes #1323

#### Checklist
- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/test/utils.ts
+++ b/packages/allure-cucumberjs/test/utils.ts
@@ -91,7 +91,7 @@ export const runCucumberInlineTest = async (
       fileExtension: ".js",
     });
   });
-  await writeFile(join(testDir, "package.json"), String.raw`{"name": "dummy"}`, "utf8");
+  await writeFile(join(testDir, "package.json"), JSON.stringify({ name: "dummy" }), "utf8");
   await step("world.js", async () => {
     await writeFile(worldFilePath, worldContent, "utf8");
     await attachment("world.js", worldContent, {
--- a/packages/allure-jest/src/environmentFactory.ts
+++ b/packages/allure-jest/src/environmentFactory.ts
@@ -20,13 +20,13 @@ import {
   getFrameworkLabel,
   getHostLabel,
   getLanguageLabel,
-  getLegacyTestCaseIdFromFullName,
   getPackageLabel,
   getPosixPath,
   getProjectName,
   getSuiteLabels,
   getThreadLabel,
   includedInTestPlan,
+  md5,
   parseTestPlan,
 } from "allure-js-commons/sdk/reporter";
 import { setGlobalTestRuntime } from "allure-js-commons/sdk/runtime";
@@ -251,7 +251,7 @@ const createJestEnvironment = <T extends typeof JestEnvironment>(Base: T): T => 
             getLanguageLabel(),
             getFrameworkLabel("jest"),
             getPackageLabel(this.testPath),
-            getFallbackTestCaseIdLabel(getLegacyTestCaseIdFromFullName(legacyFullName)),
+            getFallbackTestCaseIdLabel(md5(legacyFullName)),
             getHostLabel(),
             getThreadLabel(env.JEST_WORKER_ID),
             ...getEnvironmentLabels(),
--- a/packages/allure-jest/src/environmentFactory.ts
+++ b/packages/allure-jest/src/environmentFactory.ts
@@ -16,11 +16,14 @@ import {
   ReporterRuntime,
   createDefaultWriter,
   getEnvironmentLabels,
+  getFallbackTestCaseIdLabel,
   getFrameworkLabel,
   getHostLabel,
   getLanguageLabel,
+  getLegacyTestCaseIdFromFullName,
   getPackageLabel,
   getPosixPath,
+  getProjectName,
   getSuiteLabels,
   getThreadLabel,
   includedInTestPlan,
@@ -161,7 +164,10 @@ const createJestEnvironment = <T extends typeof JestEnvironment>(Base: T): T => 
       const newTestPath = newTestSuitePath.concat(testTitle);
       const newTestId = getTestId(newTestPath);
 
-      return `${getPosixPath(this.testPath)}#${newTestId}`;
+      const projectName = getProjectName();
+      const fullNamePath = getPosixPath(this.testPath);
+      const fullNameBase = projectName ? `${projectName}:${fullNamePath}` : fullNamePath;
+      return `${fullNameBase}#${newTestId}`;
     }
 
     #handleSuiteStart() {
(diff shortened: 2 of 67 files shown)


epszaw (changes requested):
Add tests which check the change for every integration and of course, use the function in the integrations. Currently, it isn't used anywhere except the tests for itself.

packages/allure-js-commons/test/sdk/reporter/utils.spec.ts
```diff
@@ -54,3 +55,37 @@ describe("getSuiteLabels", () => {
...
+describe("getPackageName", () => {
+  it("should cache the package name on subsequent calls", () => {
+    const first = getPackageName();
+    const second = getPackageName();
+    expect(first).toBe(second);
+    expect(first).toBe("allure-js-commons");
+  });
+
+  it("should continue searching parent directories if package.json has no name", () => {
+    // If a package.json exists but has no "name" field (or is malformed),
```
epszaw: These ai comments are pretty useless. The test name should describe well enough your check's intention. Add comments only when your solution isn't obvious and needs additional clarifications.

packages/allure-mocha/test/spec/api/testplan.test.ts
```diff
@@ -15,7 +15,8 @@ describe("testplan", () => {
       ));
     });
 
-    it("selects only one test by fullName", () => {
+    it.skip("selects only one test by fullName", () => {
```
epszaw: Don't skip failed tests, fix them once they're broken

packages/allure-js-commons/src/sdk/reporter/utils.ts
```diff
@@ -70,6 +70,24 @@ export const readImageAsBase64 = async (filePath: string): Promise<string | unde
...
+
+  try {
+    const packageJsonContent = fs.readFileSync(packageJsonPath, "utf-8");
+    const packageJson = JSON.parse(packageJsonContent);
+
+    if (packageJson.name && typeof packageJson.name === "string") {
+      return packageJson.name;
+    }
+  } catch {
+    // No package.json found or invalid
```
epszaw: Also remove the comment 🙃

epszaw (changes requested):
Tests for some packages are missing. Add tests not only for labels, but for tests results (check `fullName` directly)

delatrie (changes requested):

packages/allure-cucumberjs/test/utils.ts
```diff
@@ -12,12 +12,13 @@ type CucumberRunOptions = {
...
   env?: Record<string, string>;
   cwd?: string;
+  packageName?: string;
 };
 
 export const runCucumberInlineTest = async (
   features: string[],
   stepsDefs: string[],
-  { parallel = true, testPlan, env, cwd }: CucumberRunOptions = {},
+  { parallel = true, testPlan, env, cwd, packageName }: CucumberRunOptions = {},
```
epszaw: Better to add `files` argument instead of `packageName` (see `playwright` test utils as a reference) and write `package.json` file.

The argument also will give us more test flexibility.

epszaw (changes requested):

packages/allure-cucumberjs/test/utils.ts
```diff
@@ -12,12 +12,13 @@ type CucumberRunOptions = {
...
   env?: Record<string, string>;
   cwd?: string;
+  files?: Record<string, string | Buffer>;
 };
 
 export const runCucumberInlineTest = async (
   features: string[],
   stepsDefs: string[],
-  { parallel = true, testPlan, env, cwd }: CucumberRunOptions = {},
+  { parallel = true, testPlan, env, cwd, files }: CucumberRunOptions = {},
```
delatrie: What's the motivation for introducing `files`? It's not used anywhere.

packages/allure-jest/src/environmentFactory.ts
```diff
@@ -126,7 +128,10 @@ const createJestEnvironment = <T extends typeof JestEnvironment>(Base: T): T =>
       const newTestPath = newTestSuitePath.concat(testTitle);
       const newTestId = getTestId(newTestPath);
 
-      return `${getPosixPath(this.testPath)}#${newTestId}`;
+      const projectName = getProjectName();
+      const fullNamePath = getPosixPath(getRelativePath(this.testPath));
```
delatrie: `testPath` is already relative. Jest has its own notion of project root dir, so we use it instead of ours.

```suggestion
      const fullNamePath = getPosixPath(this.testPath);
```

packages/allure-jest/test/spec/testplan.test.ts
```diff
@@ -55,7 +35,37 @@ it("respects testplan", async () => {
...
+        const selector = (file: string, name: string) => `${packageName}:${getPosixPath(file)}#${name}`;
+        const exampleTestPlan: TestPlanV1 = {
+          version: "1.0",
+          tests: [
+            {
+              id: 1,
+              selector: selector(
+                "nested/super strange nested/super strange name.test.js",
+                "also nested should execute",
+              ),
```
delatrie: Let's just use literals. Also, let's put the test plan definition where it was and pass it to `runJestInlineTest` instead of writing it ourselves.

```suggestion
              selector: "allure-jest:nested/super strange nested/super strange name.test.js#also nested should execute",
```

delatrie (changes requested):
Additionally, we're in the process of figuring out how to make a transition from old to new IDs less painful for TestOps users. Once we come up with a solution, it will probably need to be implemented as a part of this PR (or in a follow-up PR, before the release).

packages/allure-playwright/src/index.ts
```diff
@@ -165,26 +169,32 @@ export class AllureReporter implements ReporterV2 {
     const titleMetadata = extractMetadataFromString(test.title);
     const project = suite.project()!;
     const testFilePath = path.relative(project?.testDir, test.location.file);
-    const relativeFileParts = testFilePath.split(path.sep);
-    const relativeFile = relativeFileParts.join("/");
+    const relativeFilePosix = getPosixPath(testFilePath);
+    const relativeFileParts = relativeFilePosix.split("/");
+    const projectName = getProjectName();
+    const relativeFileForId = getPosixPath(getRelativePath(test.location.file));
```
delatrie: For frameworks that support a notion of the project root, we prefer to use the framework's API rather than our own root-resolution algorithm (`getRelativePath`).

Given that, and knowing the new `testCaseIdBase` differs from the legacy one only in `projectName` in front of it, we can simplify the logic:

```ts
const relativeFileParts = testFilePath.split(path.sep);
const relativeFile = relativeFileParts.join("/");
const projectName = getProjectName();

// ...

const legacyTestCaseIdBase = `${relativeFile}#${nameSuites}${test.title}`;
const testCaseIdBase = projectName ? `${projectName}:${legacyTestCaseIdBase}` : legacyTestCaseIdBase;
```

delatrie (changes requested):

delatrie (approved):

epszaw (approved):

PR #1443, opened 2026-04-01
feat: add sync allure runtime API support across adapters

baev:
### Context

This introduces sync runtime API support across the JS adapters with a new `allure-js-commons/sync` facade. The goal is to make Allure metadata and steps usable from synchronous helpers, matcher integrations, and fully sync test code, without forcing everything through async wrappers. The facade mirrors the existing runtime API, but keeps `step()` strict-sync: the callback must finish synchronously and must not return a `Promise`.

Most of the work is in `allure-js-commons`, where the shared sync facade and runtime support were added on top of the existing runtime model instead of creating a separate runtime registration system. From there, adapters can opt into sync support using the same runtime object. This PR also adds framework-level support and tests where the runtime model can support it cleanly, including Playwright, while preserving the existing async API behavior.

```ts
import * as allure from "allure-js-commons/sync";

allure.epic("Authentication");
allure.parameter("browser", "chromium");

allure.step("Submit valid credentials", () => {
  allure.attachment("request", JSON.stringify({ login: "jane" }), {
    contentType: "application/json",
  });
});
```

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


epszaw (changes requested):
Mention sync API in rest README files (the PR does it for `allure-js-commons` and `allure-playwright` packages)

packages/allure-js-commons/src/sdk/runtime/types.ts
```diff
@@ -8,6 +8,48 @@ import type {
   StatusDetails,
 } from "../../model.js";
 
+export interface SyncStepContext {
```
delatrie, line 17: I suggest moving the `StepContext` type (an async version) here as well for the symmetry.

packages/allure-js-commons/src/sdk/runtime/runtime.ts
```diff
@@ -1,7 +1,9 @@
-import { noopRuntime } from "./NoopTestRuntime.js";
-import type { TestRuntime } from "./types.js";
+import { noopRuntime, noopSyncRuntime } from "./NoopTestRuntime.js";
+import type { SyncTestRuntime, TestRuntime } from "./types.js";
 
 const ALLURE_TEST_RUNTIME_KEY = "allureTestRuntime";
+const pwAutoconfigModuleName = "allure-playwright/autoconfig";
+const localRequire = typeof require === "function" ? require : undefined;
```
delatrie: Use `createRequire` if `require` is undefined (which happens in ESM), see [here](https://github.com/allure-framework/allure-js/blob/6b71e3bc1118c1610d21ae80921a5adee77cce1a/packages/allure-mocha/src/extraReporters.ts#L13). Otherwise, the sync API won't work in Playwright ESM.

A sidenote: The tests don't catch it because they only test the CJS version. Ideally, we should test both versions, but if we have to pick one, I would test ESM, as the world is transitioning towards it. It's only a matter of putting `package.json` with `"type": "module"` to the test dir and fixing `playwright.config.js` to use the ESM syntax.

packages/allure-mocha/test/spec/api/runtime/sync.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  it("handles sync runtime steps", () => {
+    const testResult = results.tests.find((test) => test.name === "a sync runtime step")!;
+
+    expect(testResult.labels).toContainEqual(expect.objectContaining({ name: "mode", value: "sync" }));
+    expect(testResult.steps[0]).toMatchObject({
+      name: "renamed outer",
+      status: Status.PASSED,
+      parameters: [{ name: "browser", value: "chromium" }],
+    });
```
delatrie: Since the sample defines a nested step, it should also be asserted, e.g., like this (or removed from the sample):

```js
expect(testResult.steps).toMatchObject([
  {
    name: "renamed outer",
    status: Status.PASSED,
    parameters: [{ name: "browser", value: "chromium" }],
    steps: [
      expect.objectContaining({
        name: "inner",
        status: Status.PASSED,
        steps: [
          expect.objectContaining({
            name: "foo.txt",
            attachments: [
              expect.objectContaining({
                name: "foo.txt",
                type: "text/plain",
              }),
            ],
          }),
        ],
      }),
    ],
  },
]);
```

It's also relevant for other integrations that have tests like this.

packages/allure-mocha/test/spec/api/runtime/sync.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const [attachmentStep] = testResult.steps[0].steps[0].steps;
+    const [attachment] = attachmentStep.attachments;
+    expect(attachmentStep.name).toBe("foo.txt");
+    expect(Buffer.from(results.attachments[attachment.source] as string, "base64").toString("utf8")).toBe("bar");
+  });
+
+  it("writes sync runtime globals", () => {
+    const globalsEntries = Object.entries(results.globals ?? {});
+    expect(globalsEntries.length).toBeGreaterThan(0);
```
delatrie: We can assert the exact length here:

```suggestion
    expect(globalsEntries.length).toEqual(2);
```

packages/allure-mocha/test/spec/api/runtime/sync.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const allErrors = globalsEntries.flatMap(([, info]) => info.errors);
+    const allAttachments = globalsEntries.flatMap(([, info]) => info.attachments);
+    expect(allErrors).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          message: "global setup failed",
+          trace: "stack",
+        }),
+      ]),
+    );
```
delatrie: `arrayContaining` can be removed, as we expect a single error here:

```suggestion
    expect(allErrors).toEqual([
      expect.objectContaining({
        message: "global setup failed",
        trace: "stack",
      }),
    ]);
```

packages/allure-mocha/test/spec/api/runtime/sync.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const globalAttachmentRef = allAttachments.find((attachment) => attachment.name === "global-log");
+    expect(globalAttachmentRef?.type).toBe("text/plain");
+    expect(Buffer.from(results.attachments[globalAttachmentRef!.source] as string, "base64").toString("utf-8")).toBe(
+      "hello",
+    );
```
delatrie: We can tighten the assertion because a single attachment is expected here:

```suggestion
    expect(allAttachments).toMatchObject([
      expect.objectContaining({
        name: "global-log",
        type: "text/plain",
      }),
    ]);
    expect(
      Buffer.from(results.attachments[allAttachments[0].source] as string, "base64").toString(
        "utf-8",
      ),
    ).toBe("hello");
```

packages/allure-cypress/src/browser/runtime.ts
```diff
@@ -240,17 +247,21 @@ class AllureCypressTestRuntime implements TestRuntime {
   }
 
   #enqueueMessageAsync(message: CypressMessage): PromiseLike<void> {
-    enqueueRuntimeMessage(message);
+    this.#enqueueMessageSync(message);
     return Cypress.Promise.resolve();
   }
 
+  #enqueueMessageSync(message: CypressMessage) {
```
delatrie: nit: a noop function; `enqueueRuntimeMessage` can be used directly

packages/allure-playwright/src/index.ts
```diff
@@ -662,6 +663,18 @@ export class AllureReporter implements ReporterV2 {
...
   }
 
+  #resolvePlaywrightStepStatus(step: TestStep, fallbackStatus: Status): Status {
+    const annotatedStatus = step.annotations.find((annotation) =>
+      isAllureStepStatus(annotation.description),
+    )?.description;
+
+    if (isAllureStepStatus(annotatedStatus)) {
+      return annotatedStatus;
+    }
```
delatrie: nit: if `annotatedStatus` is defined, it's guaranteed to be valid, as `isAllureStepStatus` has already checked it inside `find`.

```suggestion
    if (annotatedStatus !== undefined) {
      return annotatedStatus as Status;
    }
```

packages/allure-playwright/src/runtime.ts
```diff
@@ -54,10 +142,146 @@ export class AllurePlaywrightTestRuntime extends MessageTestRuntime {
...
+      step.complete({ error });
+      throw error;
+    }
+  }
+
+  #startSyncStep(name: string): PlaywrightInternalStep {
+    return this.#getInternalTestInfo()._addStep({
+      category: "test.step",
+      title: name,
+      infectParentStepsWithError: true,
```
delatrie: I believe this was a mistake on our part: disregarding a parent step's passed status and overwriting it with its substep's failure (see [here](https://github.com/allure-framework/allure-js/blob/690130a3a9ff1ca6c857e2939aedffa3c1123e0c/packages/allure-playwright/src/utils.ts#L89)). Similarly, we shouldn't use `infectParentStepsWithError: true` here.

Overwriting a step's status contradicts our view of steps as a logging mechanism: a step must reliably map to the code execution result. But with overwriting, these two outcomes become indistinguishable from the step status perspective:

```js
step("outer", () => {
  step("inner", () => {
    throw new Error("inner step failed!");
  });
});
```

```js
step("outer", () => {
  try {
      step("inner", () => {
        throw new Error("inner step failed!");
      });
  } catch {}
});
```

In both cases, the outer and inner steps will both be marked as `failed`.

To my understanding, allure-playwright is the only integration that exhibits this behavior. I quickly checked allure-junit5, allure-testng, allure-pytest, and allure-nunit - none of them do that.

For soft error visibility, we have a failed substeps counter in Allure Awesome. There is no need in abusing step statuses for that.

delatrie: We can improve Playwright UI/Trace by providing steps with [`location`](https://playwright.dev/docs/api/class-test#test-step-option-location). Without it, Playwright points to allure-playwright's `runtime.js` as the source of the step:

Image

<br>

By default, Playwright sets the location to the nearest frame outside playwright and playwright-core (see [here](https://github.com/microsoft/playwright/blob/9b383a201b7a36451e2d87ca83af36a44472b310/packages/playwright/src/worker/testInfo.ts#L305); also see [`captureRawStack`](https://github.com/microsoft/playwright/blob/9b383a201b7a36451e2d87ca83af36a44472b310/packages/isomorphic/stackTrace.ts#L31) and [`filteredStackTrace`](https://github.com/microsoft/playwright/blob/cb6a61be767bc730876c40dc759668345b2d9112/packages/playwright/src/util.ts#L64)). That's why `runtime.js` appears.

If we implement a similar logic that removes Allure frames from the stack and set `location` explicitly, the original `step` and `logStep` calls in the user-space will be highlighed as expected.

Allure can also benefit from it if we choose to display source locations there in the future. In any case, since this also relevant for async `step`, we can revisit this later, after this PR is done.

packages/allure-playwright/src/runtime.ts
```diff
@@ -54,10 +142,146 @@ export class AllurePlaywrightTestRuntime extends MessageTestRuntime {
...
+    }
+
+    step.complete(stepError ? { error: stepError } : {});
+  }
+
+  #stepSync<T>(name: string, body: () => T): T {
+    const step = this.#startSyncStep(name);
+
+    try {
+      const result = this.#runInStepZone(step, body);
```
delatrie: A zone in Playwright is an AsyncLocalStorage that holds the current step (and some other unrelated stuff). It extends the scope of the value to all callbacks scheduled inside the function, which, in the context of synchronous steps, may lead to unexpected behavior.:

```js
import { test, expect } from "@playwright/test";
import { step } from "allure-js-commons/sync";

test("steps", async () => {
  await test.step("Common async parent", async () => {
    await new Promise((resolve) => {
      step("Sibling 1", () => {
        setImmediate(() => {
          step("Sibling 2", () => resolve());
        });
      });
    });
  });
});
```

At runtime:

1. An async root step (defined with `test.step`) runs `Sibling 1` synchronously.
2. `Sibling 1` schedules a callback with `setImmediate` and then finishes. The callback is yet to run.
3. The root step pauses on a promise and waits for the callback to resolve it.
4. Node.js invokes the callback, which runs a sync step, `Sibling 2`.
5. `Sibling 2`, in turn, resolves the promise to resume the root step and immediately finishes.
6. The root step resumes. As there is no code left to run, the root step completes, concluding the test.

The expected step structure is:

```
- Common async parent
  - Sibling 1
  - Sibling 2
```

But Allure shows this:

Image

 <br>

And Playwright UI/Trace shows this:

Image

<br>

The reason is that by using `with("stepZone", step).run(body)`, we establish `Sibling 1` as a parent step for the duration of the callback, even though `Sibling 1` completes before the callback is invoked by Node. The UI/Trace and reporters (including Allure) show this differently but both visualizations don't match the expectations.


**A solution:**

1. Run the body directly, without involving AsyncLocalStorage (i.e., zones).
2. Use a plain variable to track the current step instead of AsyncLocalStorage:

   ```js
   let currentSyncStep: PlaywrightInternalStep | undefined;
   #stepSync<T>(name: string, body: () => T): T {
     const step = this.#startSyncStep(name, currentSyncStep);
     const previousCurrentStep = currentSyncStep;
     currentSyncStep = step;
     try {
       // ...
     } finally {
       currentSyncStep = previousCurrentStep;
     }
   }
   ```

3. Pass the variable as the second argument to `_addStep` and `_attach`.

An extra benefit would be avoiding a bunch of Playwright non-public members.

packages/allure-playwright/src/runtime.ts
```diff
@@ -1,12 +1,100 @@
...
+
+type PlaywrightInternalStep = {
+  stepId: string;
+  info: {
+    annotations: PlaywrightInternalAnnotation[];
+  };
+  complete: (result: { error?: unknown }) => void;
+};
+
+type PlaywrightInternalTestInfo = ReturnType<typeof test.info> & {
```
delatrie: > `ReturnType<typeof test.info>`

Can be simplified to `TestInfo`, a public type from `@playwright/test`, which we're actually aiming to extend here.

packages/allure-playwright/src/runtime.ts
```diff
@@ -1,12 +1,100 @@
...
-import { type RuntimeMessage } from "allure-js-commons/sdk";
+import { Status, type AttachmentOptions, type ParameterMode } from "allure-js-commons";
+import { getStatusFromError, isPromise, type RuntimeMessage } from "allure-js-commons/sdk";
 import { ALLURE_RUNTIME_MESSAGE_CONTENT_TYPE } from "allure-js-commons/sdk/reporter";
+import type { SyncTestRuntime } from "allure-js-commons/sdk/runtime";
 import { MessageTestRuntime } from "allure-js-commons/sdk/runtime";
 
+import { ALLURE_STEP_STATUS_ANNOTATION } from "./syncAnnotations.js";
+
+type PlaywrightInternalAttachment = {
```
delatrie: I suggest moving internal Playwright stuff (types and functions) into a separate file like `playwrightInternals.ts`. Having a dedicated file with a suitable name will give us the first suspect if some changes in Playwright's internal implementation break us.

delatrie (changes requested):

delatrie (approved):

epszaw (approved):

Issue #342, opened 2021-08-27
[Question] allure-playwright release cadence


pavelfeldman:
Playwright releases once a month and we are likely to add / improve allure integration with every release for several months. How should we go about pushing these updates? For example, Playwright 1.14 introduced steps and they are supported in this repo now, but users won't be able to use them until npm is published.

baev:
yeah, Allure team is agree on that. And we are so excited to support you. You can always DM me ( or even better mail to allure at qameta.io ) in case of some problems, and we'll do our best to help you ;)

I tried to release the changes this weekend, but the release pipeline is broken again for some reason. I'm so tired of all the lerna related issues, so I decided to change to yarn workspaces. The change is almost done and I expect the release later this week. 

BTW  From our side, the goal is to push allure js integrations out of beta as soon as possible. So I'll add additional focus on JS integrations next few month, and we should be good on releases

baev:
btw we all set, all the changes released, and, finally, the release flow is fully automated 🎉

PR #1446, opened 2026-04-07
added bun integration

formaceft-93:
Closing [#342](https://github.com/allure-framework/allure-org/issues/342)

Adds Bun support to allure-jest, including preload-based reporting, Bun-compatible modifier APIs, and aligned Allure results for Bun test runs.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


epszaw (changes requested):
The integration doesn't have common runtime code with jest, so better to create separate package for it.

We could keep it in the jest package in case it would be just another jest-environment compatible with bun, but it isn't.

epszaw (approved):

PR #1467, opened 2026-05-12
Add Allure Chai assertion step reporting

baev:
### Context

The integration is opt-in and does not change runner behavior automatically. It also avoids duplicate reporting for assertion APIs owned by other integrations: Vitest `expect` calls and Cypress global Chai assertions are ignored by `allure-chai`, while separately imported Chai instances still report normally.

Example output:

```ts
expect(user).to.have.property("name", "Ada");
// expect({"id":1,"name":"Ada"}).to.have.property("name", "Ada")

expect(token).to.not.be.empty;
// expect("abc").to.not.be.empty

assert.equal(actualName, "Ada");
// assert.equal("Grace", "Ada")
```

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests


packages/allure-chai/scripts/fixup.mjs
```diff
@@ -0,0 +1,38 @@
+import { mkdirSync, writeFileSync } from "node:fs";
```
vdvukhzhilov, line 1: This file looks like some leftover script 🤔

vdvukhzhilov (approved):

packages/allure-chai/scripts/fixup.mjs
baev: nope, that's the fix we need to apply if we wan't to release both esm and commonjs versions of the package

PR #1502, opened 2026-05-29
fix(cucumberjs): use scenario line in testCaseId to prevent duplicate name collision

### LINKED ISSUE #1387: cucumber.js + playwright + allure shows N-1 tests in a report
For a long time, I have been facing the issue that allure report shows N-1 tests in a report (e.g. 99 out of 100), the last test case is lost somehow
At the same time, all other available reports show ok, xml and json files contain the data for all tests

Asked AI about that, it suggested that because allure writes files asynchronously, it may need more time to flush, but adding timeouts after all tests will not help 

The test case that is lost is not the last test case. It is somewhere in the middle of the test run

I am using azure devops and one of the extensions in the marketplace to build and show a report

**To Reproduce**
Complete a lot of test cases and generate a report

**Expected behavior**
all test cases are shown

**Screenshots**

 - OS: Windows 11
 - Node 22, 24, Cucumber.js v12.x, allure-cucumberjs 3.4.1
 - tests are running in Azure DevOps agents

**Additional context**
generating reports locally produces the same results, so the configuration of the plugin should be ok

I have the following format configuration in cucumber config
format: [
        ...common.format,
        "allure-cucumberjs/reporter:allure",
        'json:./reports/cucumber.json',
        'html:./reports/cucumber-embedded.html',
        'junit:./reports/cucumber.xml'
    ],
formatOptions: {
        ...common.formatOptions,
        resultsDir: "allure-results"
    },

todti:
## Problem

Fixes #1387.

When two `Scenario` blocks in the same feature file share an identical name (but have different steps), the reporter computed the same `testCaseId` for both:

```ts
// before
const testCaseId = md5(`${fullNameBase}#${scenario?.name ?? pickle.name}`);
```

`allure-report` (the Java/HTML generator) deduplicates test results by `testCaseId`, so one of the two tests was silently dropped from the report — users saw **N-1 tests** instead of N. All other reporters (JSON, JUnit XML, HTML) were unaffected because they don't deduplicate.

## Fix

Include `scenario.location.line` in the `testCaseId` hash. Two scenarios with the same name but on different lines now get distinct ids:

```ts
// after
const scenarioLine = scenario?.location?.line ?? "";
const testCaseId = md5(`${fullNameBase}#${scenario?.name ?? pickle.name}#${scenarioLine}`);
```

**Scenario Outline rows are unaffected** — all rows of one outline share the same `Scenario Outline:` line number, so they keep the same `testCaseId`. This is required for cross-run history tracking (you want to see the history of "row 1 of this outline" across multiple runs, not treat each run as a fresh test).

## Breaking Change

`testCaseId` changes for **every** test in every feature file, because the hash input gains a `#<line>` suffix. Existing allure history tied to the old hash will be lost after upgrading — the first run after the upgrade will appear as a new test with no history.

`_fallbackTestCaseId` (the `legacyTestCaseId` used for allure2 → allure3 history migration) is **unchanged**, so the migration path from older allure versions is preserved.

## Testing

Added `duplicate-names.feature` with two `Scenario: same name` blocks and a new test asserting they produce different `testCaseId` values. The existing test for Scenario Outline (`should create same test case id for scenario outline tests with template name`) continues to pass.

### DIFF
--- a/packages/allure-cucumberjs/src/reporter.ts
+++ b/packages/allure-cucumberjs/src/reporter.ts
@@ -246,7 +246,8 @@ export default class AllureCucumberReporter extends Formatter {
     const posixPath = getPosixPathRelativeToProjectRoot(pickle);
     const fullNameBase = projectName ? `${projectName}:${posixPath}` : posixPath;
     const fullName = `${fullNameBase}#${pickle.name}`;
-    const testCaseId = md5(`${fullNameBase}#${scenario?.name ?? pickle.name}`);
+    const scenarioLine = scenario?.location?.line ?? "";
+    const testCaseId = md5(`${fullNameBase}#${scenario?.name ?? pickle.name}#${scenarioLine}`);
     const legacyTestCaseId = md5(`${posixPath}#${scenario?.name ?? pickle.name}`);
     const result: Partial<TestResult> = {
       name: pickle.name,
--- a/packages/allure-cucumberjs/test/samples/features/duplicate-names-rule.feature
+++ b/packages/allure-cucumberjs/test/samples/features/duplicate-names-rule.feature
@@ -1,0 +1,14 @@
+Feature: duplicate scenario names inside rules
+
+  Rule: first rule
+
+    Scenario: same name
+      Given a passed step
+
+    Scenario: same name
+      Given a failed step
+
+  Rule: second rule
+
+    Scenario: same name
+      Given a broken step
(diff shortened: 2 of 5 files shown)


epszaw (approved):

baev (changes requested):
the change will make testCaseId unstable. Any changes to the file will lead to the new testCaseId

baev:
we intentionally exclude line coordinate from testCaseId generation in all cucumber implementations we have

PR #1536, opened 2026-06-29
WIP: flush unfinished test results on crash

todti:
## Summary

When a test runner crashes or exits unexpectedly (uncaught exception, timeout, OOM, `process.exit()`), any test results that haven't been written to disk are lost — the Allure report ends up empty or incomplete, making it impossible to debug what went wrong.

This PR adds emergency flush of pending test results:

- **`ReporterRuntime.flushUnfinishedTests()`** — iterates over all test results still in memory, marks unfinished ones as `broken` with the crash error message and stack trace, and writes them to disk
- **`ReporterRuntime.registerProcessExitHandler()`** — registers `uncaughtException`/`unhandledRejection` listeners to capture the actual error, then flushes on `process.exit` with the captured error details
- **All integrations updated** — each reporter calls `flushUnfinishedTests()` in its end-of-run hook and `registerProcessExitHandler()` at init time

### Integrations covered

| Package | Framework hook | Process exit handler |
|---|---|---|
| allure-vitest | `onTestRunEnd` | ✅ |
| allure-playwright | `onEnd` (when status ≠ passed) | ✅ |
| allure-jest | `run_finish` | ✅ |
| allure-mocha | `done()` | ✅ |
| allure-jasmine | `jasmineDone()` | ✅ |
| allure-cucumberjs | `onTestRunFinished()` | ✅ |
| allure-cypress | `endRun()` | ✅ |
| allure-ava | already handled + exit handler | ✅ |
| allure-bun | `finishRunState()` | ✅ |
| allure-node-test | `finish()` | ✅ |
| allure-codeceptjs | via allure-mocha | via allure-mocha |

### What this does NOT cover

- `SIGKILL` / OOM killer — instant process death, no callbacks possible
- For that, incremental result writing would be needed (future work)

## Test plan

- [x] Unit tests for `flushUnfinishedTests` (7 tests covering: broken status, custom message, trace, status preservation, multiple tests, testplan skip, stopped-but-not-written)
- [x] Type-check passes for all 10 packages
- [ ] Integration tests with real test runners
- [ ] Manual verification with `allure serve`


epszaw:
It's ok to lose data if the process terminates because in case of external termination sometimes we don't see even native test framework output (depends on settings). 

Furthermore, if we run the tests using the `allure run` command, we'll see the reason for the failure and all the reported data, so the picture will be clear enough for the user – it's better to see the reason for the process termination as a global error instead of a batch of `unknown` tests. 

So, let's close the PR for now and re-open it as soon as we need it.

PR #1561, opened 2026-08-04
global errors

formaceft-93:
report failed JS integration hooks as Allure global errors while preserving their existing fixture/container output

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cucumberjs/test/spec/hooks.test.ts
+++ b/packages/allure-cucumberjs/test/spec/hooks.test.ts
@@ -47,7 +47,8 @@ it("handles passed hooks", async () => {
 });
 
 it("handles failed hooks", async () => {
-  const { tests, groups } = await runCucumberInlineTest(["hooks"], ["failedHooks"]);
+  const { tests, groups, globals } = await runCucumberInlineTest(["hooks"], ["failedHooks"]);
+  const allErrors = Object.values(globals ?? {}).flatMap((info) => info.errors);
 
   expect(tests).toHaveLength(1);
   expect(tests).toEqual(
@@ -92,6 +93,20 @@ it("handles failed hooks", async () => {
       }),
     ]),
   );
+  expect(allErrors).toEqual(
+    expect.arrayContaining([
+      expect.objectContaining({
+        message: expect.stringContaining("hook failed: Error: before error"),
+        timestamp: expect.any(Number),
+      }),
+      expect.objectContaining({
+        message: expect.stringContaining("hook failed: Error: after error"),
+        timestamp: expect.any(Number),
+      }),
+    ]),
+  );
+  expect(allErrors.filter((error) => error.message?.includes("before error"))).toHaveLength(1);
+  expect(allErrors.filter((error) => error.message?.includes("after error"))).toHaveLength(1);
 });
 
 it("handles hooks with steps", async () => {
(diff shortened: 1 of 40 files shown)


packages/allure-playwright/src/index.ts
```diff
@@ -1293,4 +1317,21 @@ export const allure = allurePlaywrightLegacyApi;
...
+  type: "global_error",
+  data: {
+    ...details,
+    message: details.message ? `${name} failed: ${details.message}` : `${name} failed`,
+  },
+});
+
+const isDirectHookStep = (step: TestStep) =>
+  (step.parent?.title === BEFORE_HOOKS_ROOT_STEP_TITLE || step.parent?.title === AFTER_HOOKS_ROOT_STEP_TITLE) &&
+  /\bhook$/i.test(step.title);
```
vdvukhzhilov, line 1330: In playwright we can have a named hook

```
test.beforeEach('Open start URL', async ({ page }) => {
  console.log(`Running ${test.info().title}`);
  await page.goto('https://my.start.url/');
});
```

I don't think that this would match it

packages/allure-cucumberjs/test/spec/hooks.test.ts
```diff
@@ -92,6 +93,20 @@ it("handles failed hooks", async () => {
...
+        timestamp: expect.any(Number),
+      }),
+      expect.objectContaining({
+        message: expect.stringContaining("hook failed: Error: after error"),
+        timestamp: expect.any(Number),
+      }),
+    ]),
+  );
+  expect(allErrors.filter((error) => error.message?.includes("before error"))).toHaveLength(1);
+  expect(allErrors.filter((error) => error.message?.includes("after error"))).toHaveLength(1);
```
vdvukhzhilov, line 109: Cucumber also has `BeforeAll` and `AfterAll`, hmmm...
https://cucumber.io/docs/cucumber/api/#global-hooks

I think we should also add global errors for them too (if it is not added atm)

vdvukhzhilov (changes requested):
Do not see anything for `allure-vitest`. Is it already working?

Also tests:

- `beforeEach` fails **and** `afterEach` fails on the same test - two globals, both fixtures still recorded
- throw with an empty message - `{hook} failed` (no extra `:`)
- nested `describe`
- not every available hook is tested to fail in all integrations

formaceft-93:
> Do not see anything for `allure-vitest`. Is it already working?
> 
> Also tests:
> 
> * `beforeEach` fails **and** `afterEach` fails on the same test - two globals, both fixtures still recorded
> * throw with an empty message - `{hook} failed` (no extra `:`)
> * nested `describe`
> * not every available hook is tested to fail in all integrations

Thanks, good catches. Vitest was not covered before, I added hook failure global errors there now. Playwright named hooks are handled via step.category === "hook" instead of matching the title, and Cucumber BeforeAll/AfterAll are covered too. For the exhaustive hook matrix, I added coverage where it exercises distinct behavior; for several fixture-based integrations the global error path is shared by all finalized failed fixtures, so I kept tests representative instead of duplicating every hook type everywhere.

vdvukhzhilov (approved):
wonderful 👍

PR #1563, opened 2026-08-10
publish result files atomically

formaceft-93:
Write Allure result files and attachments to same-directory .tmp files, fsync them, and rename them to the final names after writing completes.

This prevents live report generation and TestOps readers from observing partially written outputs, especially large attachments such as videos. Failed publications clean up their temporary files.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
+++ b/packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
@@ -1,5 +1,16 @@
-import { copyFileSync, mkdirSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
+import { randomUUID } from "node:crypto";
+import {
+  closeSync,
+  fsyncSync,
+  mkdirSync,
+  openSync,
+  readSync,
+  renameSync,
+  rmSync,
+  writeFileSync,
+  writeSync,
+} from "node:fs";
+import { basename, dirname, join } from "node:path";
 
 import type { Globals, TestResult, TestResultContainer } from "../../../model.js";
 import type { Category, EnvironmentInfo } from "../../types.js";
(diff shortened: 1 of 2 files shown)


packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
```diff
@@ -1,13 +1,83 @@
...
 import { stringifyEnvInfo } from "../utils/envInfo.js";
 
 const writeJson = (path: string, data: unknown): void => {
-  writeFileSync(path, JSON.stringify(data), "utf-8");
+  writeFile(path, JSON.stringify(data), "utf-8");
+};
+
+const buildTempPath = (path: string): string => {
+  return join(dirname(path), `${basename(path)}.${process.pid}.${randomUUID()}.tmp`);
+};
```
vdvukhzhilov: Other integrations use only `.allure-write-${uuid}.tmp` to prevent other external processes to by some regexps (like `*-result.json` as an example)

packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
```diff
@@ -1,13 +1,83 @@
...
+
+          while (bytesWritten < bytesRead) {
+            bytesWritten += writeSync(toFd, buffer, bytesWritten, bytesRead - bytesWritten);
+          }
+        }
+      });
+    } finally {
+      closeSync(fromFd);
+    }
+  });
```
vdvukhzhilov, line 95: _Not blocking._ This copies the file by hand in small chunks. Node already has a faster way to do that: `copyFileSync`.

Easier version would be to copy to the temp file with `copyFileSync`, then open that temp file and flush it to disk (`fsyncSync`), then rename it to the final name.

Also also when open the temp file for the flush, use `r+`. Do not use `"w"` cause that wipes the file you just copied 😄

packages/allure-js-commons/test/sdk/reporter/writer/FileSystemWriter.spec.ts
vdvukhzhilov, line 1: Tests only write each file once.
In real runs we often write `categories.json` and `environment.properties` again with new content. There is no check that the second write replaces the first, and that no `.tmp` file is left behind.
Java/Dart integrations cover that. A small test like this would help:

_write categories once, write them again with different data, then check the file has the second version and the folder has no leftover temp files.
Same idea for `environment.properties`._

This would be great to add 👍

packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
```diff
@@ -1,13 +1,83 @@
...
+const publishFile = (path: string, writeTempFile: (path: string) => void): void => {
+  const tempPath = buildTempPath(path);
+
+  try {
+    writeTempFile(tempPath);
+    renameSync(tempPath, path);
+  } finally {
+    rmSync(tempPath, {
+      force: true,
+    });
```
vdvukhzhilov: _Not blocking._ `force: true` only ignores `file not found`.

If deleting the temp file fails for another reason (`file locked`), that error can replace the original write/rename error. Then the logs show the cleanup problem, not the real one.

A small try/catch around `rmSync` would keep the original error. E.g. Java integration logs a warning and does the same.

vdvukhzhilov (changes requested):

packages/allure-js-commons/src/sdk/reporter/writer/FileSystemWriter.ts
formaceft-93: Good point. I kept this as-is for now to avoid the read-only attachment case: `copyFileSync` may preserve read-only permissions, and then reopening the temp file with `r+` for fsync can fail. Since we need to fsync before publishing, the manual fd copy keeps that path predictable. There’s a regression test covering read-only path attachments.

formaceft-93: Small note on the `w` flag: in the current implementation we open the temp file with `w` before writing/copying into it, so nothing already copied can be truncated. The `r+` concern applies to the alternative `copyFileSync -> reopen temp -> fsync` flow, where reopening with `w` would wipe the copied content.

vdvukhzhilov (approved):
👍

PR #1568, opened 2026-08-21
Add an option to turn off the steps generated from Cypress commands (fixes #1129)

### LINKED ISSUE #1129: Ability to turn off automatic steps and substeps creation for Cypress
It will be useful to have ability to turn off automatic steps and substeps creation for Cypress, and be able to use only those coded with allure.step

dylanpulver:
### Context

Fixes #1129.

The switch is `enabled` inside the existing `stepsFromCommands` group, next to `maxArgumentLength` and `maxArgumentDepth`, so every setting for the command log steps stays in one place. A top level key was the alternative, and it would have split the group from the flag that turns the whole group into a no-op.

```js
allureCypress(on, config, {
  stepsFromCommands: { enabled: false },
});
```

The gate sits in `shouldCreateStepFromCommandLogEntry`, the single predicate the `log:added` handler consults before anything reaches the step stack. Nothing is pushed, so nothing needs unwinding when the option is off. Default is `true`, so a config that omits the key reports exactly what it reports today.

Limits worth knowing: the flag only covers steps built from the command log. Steps from the Allure API and attachments such as screenshots come from their own handlers and keep working. The value applies to the whole spec, there is no per test override.

Three cases in `packages/allure-cypress/test/spec/commands.test.ts`, each running a spec that puts one command inside `allure.step` and one outside it. Reverting the source change fails the `enabled: false` case and leaves the other two green; those two are there to pin the default.

#### Checklist

- [ ] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-cypress/src/browser/commandLog.ts
+++ b/packages/allure-cypress/src/browser/commandLog.ts
@@ -4,11 +4,16 @@ import type { CypressLogEntry, LogStepDescriptor } from "../types.js";
 import { isDefined } from "../utils.js";
 import { reportStepStart } from "./lifecycle.js";
 import serializePropValue from "./serialize.js";
-import { getCurrentStep, getStepStack, pushStep, setupStepFinalization } from "./state.js";
+import { getConfig, getCurrentStep, getStepStack, pushStep, setupStepFinalization } from "./state.js";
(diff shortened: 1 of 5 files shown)


epszaw (approved):

epszaw:
@dylanpulver thank you for the useful contribution!

PR #1581, opened 2026-09-04
fix(codeceptjs): preserve attachments for failed before hooks

### LINKED ISSUE #1580: Attachments are missing when a test fails in a Before hook
### What happened?

Failure screenshot and Playwright trace are not attached to it if test fails in before hook. CodeceptJS saves both into the output folder, yet nothing shows up in the Allure report - not even an `*-attachment.*` file is written to the results folder, so the attachment seems to be dropped because the test result is already finished when the hook failure is handled.

The same happens for attachments added from a custom helper `_failed` hook or from an `event.test.failed` listener. If the failure happens inside the `Scenario` instead of the `Before` hook, both attachments are attached correctly.

### How can we reproduce it?

Create a suite with a failing Before hook, with the screenshot plugin enabled and the Playwright helper configured with trace: `'retain-on-failure'.`

```
Feature('First suite').tag('@debug');

Before(({ I }) => {
  I.amOnPage('/');
  I.waitForElement('.does-not-exist', 1);
});

Scenario('First scenario', ({ I }) => {
  I.dontSeeElement('.does-not-exist');
});
```

Then:
Run all suites.
After run finishes, serve Allure report.
Open the report and the failed test case.

### What did you expect?

The failure screenshot and the trace of a scenario that failed in a Before hook should be attached to the test case, the same way they are when the scenario itself fails.

### Environment
- allure-codeceptjs: 3.11.1
- CodeceptJS: 4.1.0
- playwright-core
- GitHub Actions

### Additional context

_No response_

formaceft-93:
Fixes https://github.com/allure-framework/allure-js/issues/1580

The reporter now starts the Allure test before other `test.failed` listeners run, keeps the failed test open until queued failure artifacts are available, and finalizes each pending scenario by its own uuid. This also handles multi-scenario features without mixing artifacts between results.

#### Checklist

- [x] [Sign Allure CLA][cla]
- [x] Provide unit tests

### DIFF
--- a/packages/allure-codeceptjs/src/reporter.ts
+++ b/packages/allure-codeceptjs/src/reporter.ts
@@ -14,9 +14,12 @@ interface MetaStep {
 }
 
 interface CodeceptTestWithArtifacts {
-  artifacts?: {
-    screenshot?: string;
-  };
+  artifacts?: unknown;
+}
+
+interface PendingFailedBeforeEachHookTest {
+  test: Mocha.Test;
+  uuid: string;
 }
 
 const MAX_META_STEP_NESTING = 10;
@@ -53,7 +56,8 @@ const getCodeceptStatusFromError = (error: Partial<Error>, hookName?: string): S
   return isKnownCodeceptVerificationError(error) ? Status.FAILED : Status.BROKEN;
 };
 
-const isBeforeHook = (hookName?: string) => hookName === "Before" || hookName === "BeforeSuite";
+const isBeforeEachHook = (hookName?: string) => hookName === "Before";
+const isBeforeHook = (hookName?: string) => isBeforeEachHook(hookName) || hookName === "BeforeSuite";
 
 const isErrorInstance = (value: unknown): value is Error =>
   value instanceof Error ||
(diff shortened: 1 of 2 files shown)


packages/allure-codeceptjs/src/reporter.ts
```diff
@@ -112,7 +132,18 @@ export class AllureCodeceptJsReporter extends AllureMochaReporter {
     });
   }
 
+  beforeHookTestFailed(test: Mocha.Test, error: Error, hookName?: string) {
+    if (isBeforeEachHook(hookName)) {
+      this.flushPendingFailedBeforeEachHookTest();
```
epszaw: Async `_failed` attachments can still disappear or land on the next scenario because `currentTest` changes while failure callbacks await I/O. Queue restoration of the captured test UUID and hook name before those callbacks, and move `currentTest` clearing from `flushPendingFailedBeforeEachHookTest` to the path that switches failed tests. Add a regression covering awaited I/O with one and two scenarios.

epszaw (changes requested):

epszaw (approved):

